PR #4, opened 2026-07-16
spike(tabs): native NSWindow tabbing (Ghostty-style) — Route A prototype

jiweiyuan:
## What this is

A **throwaway spike** (not for merge) pricing **Route A** for session tabs: real
native macOS window tabbing like Ghostty, instead of a custom title-bar strip.

Each ⌘T / native `+` creates a separate `NSWindow` and merges it with
`addTabbedWindow`, so macOS draws its own `NSTabBar` — the exact widget Ghostty
relocates into its titlebar.

## What it demonstrates

- The native tab-bar chrome for free: drag-reorder, compression + overflow `»`
  menu, ⌃⇥ cycling, ⌘⇧\ show-all-tabs.
- The unavoidable cost: **every tab is a whole window, so each carries its own
  sidebar** (collapse one tab's sidebar — the others keep theirs).

## What it fakes (the real rewrite)

Content still follows the store's single **global** `selectedSessionID`, so all
tabs show the same session. True per-tab content needs selection to become
**per-window**, and native tabs **can't nest into Safari-style tab groups**
(they're flat). That tradeoff is exactly what this spike measures.

## Files

- `Sources/termio/NativeTabsSpike.swift` — tabbing setup + per-window content/toolbar build
- `Sources/termio/App.swift` — enable tabbing, repoint ⌘T → New Tab, relax a few
  private members so the spike extension can reach them

## Decision

Recommendation from the exploration: **keep the custom strip**, not Route A —
native tabs drop tab-groups and force a per-window-selection rewrite for a bar
that compresses rather than scrolls anyway. This branch preserves the prototype
for reference.

### DIFF
--- a/Sources/termio/App.swift
+++ b/Sources/termio/App.swift
@@ -29,9 +29,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
     /// elsewhere in the app (`TerminalPane` filters key-window notifications with it,
     /// so the settings window never triggers the terminal refocus rescue).
     static let mainWindowFrameAutosaveName = "TermioMainWindow"
-    private var window: NSWindow!
-    private let settings = AppSettings()
-    private lazy var store = TermioStore.restored(settings: settings)
+    var window: NSWindow! // SPIKE: relaxed from private so NativeTabsSpike.swift can opt it into tabbing.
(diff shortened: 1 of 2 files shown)


jiweiyuan:
Closing: window tabs were dropped in favor of split panes; this Route-A spike isn't a direction we're pursuing.

PR #9, opened 2026-07-18
feat(terminal): cmd-click a file path inside a TUI to open it

jiweiyuan:
## Problem

When Claude Code (or any full-screen TUI) prints a file path like `docs/foo.md`, it isn't cmd-clickable in termio — even though the same path is clickable in a plain shell and in VS Code.

**Root cause (confirmed in ghostty source):** a mouse-capturing TUI turns on mouse reporting (DECSET 1002/1003/1006). Ghostty's `Surface.zig` gates link-hover on `mouse_event == .none`, so once the app grabs the mouse ghostty **stops firing its `MOUSE_OVER_LINK` action entirely**. That leaves `TerminalLinkState.hoveredURL` nil, so the existing cmd-click handler had nothing to open. (Refs: ghostty discussion #5800, pi-powerline-footer #93.)

libghostty exposes no per-cell link/OSC-8 query, so we reconstruct the path from the grid text ourselves.

## Approach

Extend the existing `linkClickMonitor` (`TermioStore.swift`) with a fallback that runs only when there's no ghostty-hovered link:

1. Resolve the clicked surface by geometry (same pattern as `TerminalContextMenu`).
2. Map the click pixel → grid cell using `TerminalViewState.surfaceSize` (cell px) + `backingScaleFactor`, accounting for grid padding and the AppKit Y-flip.
3. Read that row via the surface's `InMemoryTerminalSession.readViewportText()` (cached in a new `surfaceSessions` map, since the wrapper doesn't expose the backend session).
4. Extract the path token under the click (separator split; strip git `a/`,`b/` prefix, trailing sentence punctuation, and `:line[:col]`).
5. Resolve against the surface's OSC 7 `workingDirectory` (fallback: session worktree / project root), and open editable via `openFileInEditor(_:at:)` **only if it's a real file**. Otherwise the click passes through untouched.

The cmd-click is intercepted at the AppKit layer *before* ghostty forwards it to the TUI, which is why this works while the app has grabbed the mouse.

## Scope

- **Click-only (v1):** no hover cursor or underline. Ghostty won't render a decoration for a match it didn't make and won't run hover detection under capture, so an underline would need a cell-aligned overlay — deferred.
- **Opens editable** in the editor pane (per product choice).
- Bonus: also makes bare relative paths clickable in a plain shell (ghostty doesn't OSC-8 those either).

## Files

- `Sources/termio/TerminalPathClick.swift` (new) — the detector + cell/token/resolve helpers.
- `TermioStore.swift` — `surfaceSessions` cache + fallback wired into `linkClickMonitor`.
- `TermioStore+TerminalSurface.swift` / `TermioStore+ProjectActions.swift` — populate/tear down the cache.

## Verification

- `swift build` passes clean.
- Manual (recommended before merge): rebuild dev, have Claude Code print an existing project path, cmd-click it → opens editable at the right file (and jumps to `:line` if present); ordinary prose cmd-clicks do nothing; plain-shell URL cmd-click still opens the browser.

> Note: coordinate mapping (Y-flip / padding / CJK column drift) is the main runtime risk; a mis-mapped click simply no-ops thanks to the file-existence guard. A later precision upgrade could do a single-row `ghostty_surface_read_text` via the raw surface handle.


jiweiyuan:
Closing: superseded by the tui-path-click v2 work, which has landed in main.

PR #10, opened 2026-07-18
feat(sidebar): add Fixed project sort order and expose sort in Settings

aimen08:
## Problem

The sidebar's default (and only meaningful) sort was **Recent Activity**, which floats the active project to the top. So selecting or switching a project reshuffled the whole list on every focus — the order was never stable.

## Fix

- Add a **Fixed** sort order that keeps projects in the order they were added — nothing floats on activity or selection — and make it the new default.
- Expose the sort as a picker in **Settings › Interface**, kept in sync with the sidebar's existing sort menu.
- **Recent Activity** and **Name** remain available.

image

### DIFF
--- a/Sources/termio/Settings/InterfaceSettingsTab.swift
+++ b/Sources/termio/Settings/InterfaceSettingsTab.swift
@@ -41,6 +41,19 @@ struct InterfaceSettingsTab: View {
             } header: {
                 SectionHeaderLabel(title: "Density")
             }
+            Section {
+                Picker("Sort", selection: $settings.projectSortOrder) {
+                    ForEach(ProjectSortOrder.allCases) { order in
+                        Text(order.displayName).tag(order)
+                    }
+                }
+            } header: {
+                SectionHeaderLabel(title: "Project order")
+            } footer: {
+                Text("How the sidebar orders projects (pinned always first). Fixed holds the order you added them; Recent Activity floats the active project to the top.")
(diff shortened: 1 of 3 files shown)


jiweiyuan:
Thanks for this @aimen08 — the problem writeup is spot on. Recent Activity floating the active project on every focus was a real annoyance and killed any spatial memory of the list.

After thinking it through I'm going to go a slightly simpler route: just change the **default** sort from Recent Activity to **Name** (#16), rather than adding a new `Fixed` mode. Two reasons:

- `Fixed` sorts by the persisted `projects` array order, but our worktree/discovery reconciliation can reorder that array on focus/git changes — so `Fixed` could quietly reshuffle again, and `Project` has no `createdAt` to anchor it. `Name` is stable with zero new state.
- `Name` is also more predictable — you can find a project by first letter.

The idea you're really reaching for — a **stable, user-controlled order** — is worth doing properly later as an explicit per-project order index + drag-to-reorder (the Xcode/Finder/Dock model), which I'd rather build as its own change than approximate with insertion order.

So I'll close this in favor of #16, but genuinely appreciate you surfacing the issue and the clean PR — hope you'll send more. 🙏

aimen08:
@jiweiyuan No, I totally understand and I agree. I'd rather have something planned like you mentioned than rush for something that only patches the current issue. And don't worry, I'll send more PRs.

PR #13, opened 2026-07-18
feat(files): add open-state folder icons

orriduck:
## Summary
- show the shared Hugeicons folder pair in the Files tree
- track native outline expansion state for both row clicks and disclosure arrows
- reset expansion state when the outline tree is recreated

## Verification
- `TERMIO_CHANNEL=dev ./scripts/build-app.sh`
- launched `termio-dev.app` and confirmed the process remains running
- `git diff --check`


jiweiyuan:
Closing ahead of the app-wide Huge Icons sweep, which replaces the icon system this builds on. Open-state folder icons can be redone on top of the new set if still wanted.

PR #19, opened 2026-07-19
feat(agents): promote working status from sustained screen activity

jiweiyuan:
## Problem

A missed `working` hook left a session's sidebar dot stuck at calm while output streamed. `noteOutputActivity` had a one-way guard — it could **sustain** an existing `.working` but never **start** one — so a single lost hook report (failed hook script, early `Stop` on a continued turn) had no self-healing path. Today's stale-CLI outage made the whole class visible at once.

## Fix

Let the existing 1s screen tap promote an agent row back to `.working` when repaint evidence is sustained, behind four guards:

1. **Agent sessions only** — same `effectiveAgent != .terminal` check the hook path uses; a bare terminal never spins.
2. **Never over `needsAttention`** — a blocked permission prompt repaints too; hooks keep sole authority over attention.
3. **Input-echo window** — `PTYProcess.write` now timestamps every stdin write (one choke point covers Mac keyboard, phone companion, and `sessions send`); pokes within 1.5s of input don't count, so typing a prompt can't fake a turn.
4. **Streak threshold** — three consecutive changed-screen pokes (~3s of real streaming); a one-off redraw or resize is not a turn, and a >5s gap breaks the run.

Downgrades are untouched: only hooks and `sweepStaleWorking` ever clear the spinner, so hooks remain the one authority for `attention`/`done` (one-authority-per-pane holds).

## Known edges (accepted)

- A TUI that *continuously animates while idle* would false-promote — no built-in agent does this today (Claude/Codex idle screens are static).
- Viewport-follows-scrollback staleness is pre-existing (documented in `TermioStore+TerminalSurface.swift`) and orthogonal; scroll alone can't trigger the tap since the sink only fires on PTY output.

## Testing

- `swift build` clean.
- Manual: hooks disabled → start a Claude turn → row spins within ~4s of streaming; typing a long prompt does not spin; blocked permission prompt stays `needsAttention`.

### DIFF
--- a/Sources/termio/PTYProcess.swift
+++ b/Sources/termio/PTYProcess.swift
@@ -465,6 +465,7 @@ final class PTYProcess: @unchecked Sendable {
         writeLock.lock()
         defer { writeLock.unlock() }
         guard !writerTornDown else { return }
+        lastInputAtLocked = Date()
         // Once a backlog exists every write must append behind it, or these
         // bytes would overtake the queued remainder and reorder the stream.
         guard pendingWrite.isEmpty else {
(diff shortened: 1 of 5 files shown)


jiweiyuan:
Superseded by #20 (better-calibrated promotion: hook-quiet window, launch grace, byte-floor sustain, done preserved) + #21 (delivery layer). The one piece #19 had that #20 lacked — the PTY-level input timestamp covering phone input — was ported into #20 as d63b00f.

PR #26, opened 2026-07-20
fix(agents): make Grok resume its prior session instead of erroring

jiweiyuan:
## Problem

Relaunching a Grok session failed immediately with `Error: Session ID <uuid> is already in use. Process exited.` — Grok effectively couldn't resume.

## Root cause

Grok is configured `resume: "claude"` in its manifest, which is *correct*: Grok's CLI shares Claude's flag shape — `--session-id <uuid>` creates a session (and errors if the id already exists), `-r/--resume <uuid>` continues one.

termio switches between those two using `ResumeContext.pinnedConversationExists`. But that probe was hardcoded:

```swift
pinnedConversationExists: agent == .claudeCode
    && resumeID.map(ClaudeConversation.exists) == true
```

For Grok it is always `false` (the check only globs `~/.claude/projects`), so termio kept emitting `--session-id` on every relaunch — for an id Grok had already persisted under `~/.grok/sessions`. Hence "already in use."

## Fix

- Add a `GrokConversation.exists(id:)` probe. Grok stores each conversation as a directory named by its session UUID at `~/.grok/sessions/<url-encoded-cwd>/<id>/`; we glob the cwd folders for `<id>` rather than reconstruct Grok's percent-encoding (mirrors how `ClaudeConversation` globs `~/.claude/projects`).
- Make the `pinnedConversationExists` computation agent-aware (Claude → its store, Grok → its store).

First launch still creates (`--session-id`); every relaunch now resumes (`--resume`).

## Verification

- `swift build` passes.
- The probe returns `true` for a real on-disk session UUID and `false` for an unknown id.

### DIFF
--- a/Sources/termio/AgentDefinition.swift
+++ b/Sources/termio/AgentDefinition.swift
@@ -165,9 +165,9 @@ struct AgentDefinition: Identifiable {
         var resumeID: String
         /// Whether this session's agent has been launched in a prior app run.
         var launchedBefore: Bool
-        /// Whether Claude Code already has a saved conversation under `resumeID`.
-        /// Resuming one that doesn't exist errors, so a pinned-but-never-used session
-        /// is (re)created with `--session-id` instead.
+        /// Whether the agent already has a saved conversation under `resumeID` (Claude
+        /// Code, Grok). Resuming one that doesn't exist errors, so a pinned-but-never-used
+        /// session is (re)created with `--session-id` instead.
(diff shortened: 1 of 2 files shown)


jiweiyuan:
Superseded by #27, which makes resume config-driven — the generic SessionStore probe there subsumes this agent-specific GrokConversation fix (Grok resume is fixed as a consequence).

PR #67, opened 2026-07-24
feat(cli): send --wait — block until the turn settles, return the transcript cursor range

jiweiyuan:
One `send --wait` replaces the send-then-poll-`list` loop: it waits for the session's status to rest (done / needs-you) or the timeout, then reports the turn outcome plus the exact transcript range (`cursor`..`cursor_end`) to read the reply from. Falls back to viewport-frame settling for plain terminals with no status signal.

Recovered from an uncommitted worktree; branch is behind main and will need a rebase before merge.


jiweiyuan:
Superseded by #72, which reimplements send --wait cleanly against main's current session-control code. This branch patched the old SessionControl.swift, which main has since deleted/restructured for the @id handle system, so it could no longer be merged as-is.

PR #69, opened 2026-07-24
fix(agents): map kimi/grok hook events to attention/working status

jiweiyuan:
Notification and PermissionRequest raise attention; PermissionResult returns to working — so kimi/grok surface the same status transitions as the other hook agents.

Recovered from an uncommitted worktree; branch is behind main and will need a rebase before merge.


jiweiyuan:
Closing as superseded: main has since rewritten TermioStore+AgentStatus.swift into a generic status engine (handles attention/working/done/idle centrally), and grok.json/kimi.json already carry the hook-event + title-based attention mappings. This branch's small delta would conflict with — and risk regressing — the newer engine.

PR #72, opened 2026-07-24
feat(cli): send --wait — block until the turn settles, return the transcript range

jiweiyuan:
Reimplements #67 cleanly against main's current session-control code (the original branch patched `SessionControl.swift`, which main has since deleted/restructured for the `@id` handle system).

## What
`termio sessions send --wait <handle> "<prompt>"` (and `answer --wait`) block until the turn settles, then reply with the final status + the exact transcript line range (`cursor`..`cursor_end`) to read — replacing the send-then-poll-`list` loop with one call. `--timeout <ms>` caps the wait (clamped 1–600s; implies `--wait`).

## How
- Completion is read off the **resting status**: once a session seen `working` drops back off it for ~1.5s (or hits `needs-you`, which short-circuits). Immune to a TUI footer that keeps animating after the reply.
- Plain terminals with no status signal fall back to the screen changing and then going still.
- Reuses the `readViewportText` settling pattern `waitForBootSettle` already relies on; adds a `viewportText` helper (also surfaces the on-screen prompt when the turn ends in `needs-you`).
- Wire: `ControlRequest` gains `wait`/`timeout_ms`; CLI gains `--wait`/`--timeout`. Non-wait sends are byte-for-byte unchanged.

Builds clean (`swift build`). Not yet interactively verified against a live agent.

### DIFF
--- a/Sources/termio/Agents/SessionControl.swift
+++ b/Sources/termio/Agents/SessionControl.swift
@@ -26,14 +26,22 @@ struct ControlRequest: Decodable {
     let text: String?
     let lines: Int?
     let agent: String?
+    /// `send`/`answer` only: block until the turn the prompt kicked off settles
+    /// (or the timeout elapses) and report the outcome, instead of replying the
+    /// instant the keystrokes are delivered.
+    let wait: Bool?
+    /// The `--wait` cap in milliseconds; clamped server-side. Nil uses the default.
+    let timeoutMs: Int?
 
     private enum CodingKeys: String, CodingKey {
-        case op, format, target, text, lines, agent
+        case op, format, target, text, lines, agent, wait
         case callerSession = "caller_session"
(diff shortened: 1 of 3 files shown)


jiweiyuan:
Superseded by the Sessions CLI v2 track (docs/design/sessions-cli-v2.md §4.4). #80 + #86 rewrote all three files this PR touches (spawn is now non-blocking with detached delivery, request_once gained fail-loud + timeouts, the listener gained a streaming path), so this no longer merges cleanly. The valuable core — resting-status completion detection (seen-working → off-working ~1.5s settle, needs-you short-circuit, plain-terminal screen-still fallback) — is inherited by the Phase 3 reimplementation: send --wait and spawn --wait with one reply shape, plus actionable needs-you/done payloads.

PR #84, opened 2026-07-24
perf(git): stop the git subprocess storm behind branch/worktree watching

jiweiyuan:
## Problem

The machine ran hot with agent fleets active. `sample` + child-process sampling of the live app showed a continuous storm of short-lived `git` subprocesses (100 samples over 15s caught 104 zombie `git` children), with `tccd` at 66% and `trustd` at 31% revalidating every spawn.

Three compounding causes:

1. **BranchModel** watched the whole `.git` directory — whose busiest file is `index`, rewritten by every `git status` run by *anyone* (agents, the git pane itself) — and answered each event by spawning 2–3 `git rev-parse` processes.
2. Every event also fired an **unconditional worktree reconcile across all folder projects**: N × `git worktree list` for one repo's index refresh.
3. **The git pane's FSEvents watch** covered the whole worktree including `.build` (an agent's `swift build` = refresh storm, 4 spawns per pass), and kept refreshing while the inspector was **collapsed** — the split item keeps the hosted view (and its `@StateObject` model) alive, confirmed live: `git status --porcelain=v2` passes with no pane on screen.

## Fix

- BranchModel reads `HEAD` + the `worktrees` listing **in-process** (zero spawns per event) and reports downstream only when HEAD or the linked-worktree set actually changed. Its one remaining spawn (HEAD-dir resolution, once per watch) moves off the main thread — closing the recorded BranchModel launch-freeze path too.
- Worktree reconcile is scoped to the project owning the changed folder; app activation still re-scans everything.
- The git pane drops FSEvents from build-product dirs (`.build`, `node_modules`, …) and parks auto-refresh while the inspector is collapsed (new `store.inspectorVisible`, KVO-mirrored from the split item); deferred refreshes replay on next show.

## Verification

- `swift build` clean.
- Before: 40 ps samples in 10s → 19 caught live git children; argv showed `status --porcelain=v2`, `diff --numstat --cached`, `rev-parse --is-inside-work-tree` under the app with the inspector collapsed.
- After: pending a rebuilt dev-channel run under the same agent load (kept the release app untouched for comparison).

### DIFF
--- a/Sources/termio/App/App.swift
+++ b/Sources/termio/App/App.swift
@@ -69,6 +69,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
     // KVO on the sidebar's collapse state, so every collapse path (toolbar toggle, View menu,
     // divider drag) empties/refills the sidebar's toolbar region (see `setNavigatorItemsVisible`).
     private var sidebarCollapseObserver: NSKeyValueObservation?
+    // KVO on the inspector's collapse state, mirrored onto `store.inspectorVisible` so
+    // hosted panes (the git pane's auto-refresh) can stand down while hidden.
+    private var inspectorCollapseObserver: NSKeyValueObservation?
     // KVO on the window's effective appearance, so a *system-driven* light↔dark flip (macOS auto
     // day/night, Control Center) re-resolves the window background. In `.system` appearance mode
     // nothing else re-runs `applyWindowTransparency` on an OS flip — the settings never change —
@@ -162,6 +165,19 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
         // divider drag). No `.initial`: the launch-time sync below runs after the autosave restore.
         sidebarCollapseObserver = sidebarSplitItem?.observe(\.isCollapsed, options: [.new]) { [weak self] item, _ in
             MainActor.assumeIsolated { self?.setNavigatorItemsVisible(!item.isCollapsed) }
+        }
+        // Mirror the inspector's live collapse state onto the store, so panes it hosts
+        // can idle while hidden (a collapsed item keeps its view hierarchy alive — the
+        // git pane's watcher would otherwise keep spawning `git status` unseen). KVO
+        // catches every collapse path; `.initial` seeds the restored autosave state.
+        inspectorCollapseObserver = filesInspectorItem?.observe(
+            \.isCollapsed, options: [.initial, .new]
+        ) { [weak self] item, _ in
+            let visible = !item.isCollapsed
+            MainActor.assumeIsolated {
+                guard let store = self?.store, store.inspectorVisible != visible else { return }
+                store.inspectorVisible = visible
+            }
         }
         // Re-resolve the window background whenever the OS flips light↔dark under `.system` mode
         // (see `appearanceObserver`). Pinned Light/Dark modes never see the effective appearance
--- a/Sources/termio/FileBrowser/FileBrowserView.swift
+++ b/Sources/termio/FileBrowser/FileBrowserView.swift
@@ -65,8 +65,15 @@ struct FileBrowserView: View {
                 if let repoRoot = projectPath {
                     // Fresh identity per repo, so the panel model (selection, draft message,
                     // PR status) resets cleanly when the selected project moves.
-                    GitChangesView(repoRoot: repoRoot, changeCount: $store.gitChangeCount)
-                        .id(repoRoot)
+                    // The visibility closure reads the store live (weakly — the model may
+                    // outlive a closing window), so a collapsed inspector parks the pane's
+                    // auto-refresh even while this view stays in the hierarchy.
+                    GitChangesView(
+                        repoRoot: repoRoot,
+                        changeCount: $store.gitChangeCount,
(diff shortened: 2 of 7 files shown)


jiweiyuan:
## Benchmark results (A/B, same synthetic workload)

Harness: `scripts/bench-git-spawns.sh` (added in a199e4e) — 45s of simulated agent activity in a watched scratch repo (file appends every 0.25s, `git status` every 0.75s, a real commit every ~4s), with **exact** spawn counting via `GIT_TRACE2_EVENT` so every git the app spawns self-reports its argv.

| | main (baseline) | this branch |
|---|---|---|
| git spawns in 45s | **515** (~11.4/s) | **51** (launch-time one-offs only) |
| … `rev-parse` on the busy repo | 103 | 0 steady-state |
| … `worktree list` × unrelated projects | 51 passes × 6 projects | 1 × 2 projects (initial scan) |
| sampler git-child sightings (60ms) | 109 | 0 |

The baseline trace reproduces both amplifiers exactly as diagnosed: 2–3 `rev-parse` per index event, and every event fanning out to a `git worktree list` on **every** folder project. On the fix branch the steady-state spawn rate under load is zero; the 51 remaining are launch-time HEAD-dir resolution (one per watched folder) plus the initial full reconcile.

App-CPU% over the 45s window was too noisy to compare (session-restore work dominates a fresh launch); the spawn counts — and their tccd/trustd knock-on — are the meaningful metric.

jiweiyuan:
Ran the pending after-verification with the shipped harness (`scripts/bench-git-spawns.sh`, 45s synthetic agent workload on a scratch repo, exact counts via `GIT_TRACE2_EVENT`, same machine, same open project set):

| | main build (before) | this branch (after) |
| --- | --- | --- |
| total app-spawned `git` processes | **2,577** (~57/s) | **802** (−69%) |
| scratch-repo watcher spawns (`rev-parse`+`diff`+`status`) | 1,689 | 189 (−89%) |
| `worktree list` per project | 45 (every event) | 4 (launch + activations) |
| coarse git-child sightings (404–416 ps samples) | 194 | 87 |

The residual spawns concentrate on the main termio checkout, which had five live worktree sessions sharing its `.git` during the run — real concurrent activity, not watcher churn. Before-trace argv breakdown confirmed all three claimed causes (852 `rev-parse` on the scratch repo from BranchModel, 45×N global `worktree list`, 557 `diff` from the pane).

jiweiyuan:
Codex review of `b32b15c` (independent sibling session; no code changed). Verdict: **FIX-FIRST** — 2 bugs, 3 risks:

1. **bug** — BranchModel's change fingerprint misses real linked-worktree updates: for a linked checkout `headDirectory` is `.git/worktrees/<id>` whose local `worktrees` child is always absent (the real list is via `commondir`), so e.g. `git worktree move` never reads as changed and no reconcile fires until app reactivation. `BranchModel.swift:144,176,181,207`, `TermioStore.swift:525`
2. **bug** — the FSEvents exclusion drops any event whose path contains a component named `.build`/`node_modules`/… without checking tracked/ignored status, and returns *before* recording a deferred refresh — a repo sitting below such a dir goes permanently deaf, and reopening the inspector doesn't repair the stale list. `GitPanelModel.swift:49,118`
3. **risk** — scoped reconcile assumes path ownership = repo ownership: a primary checkout and a separately-opened linked checkout share one common git dir, but only the matching project reconciles; the sibling stays stale until a full activation scan. `TermioStore.swift:589,605`
4. **risk** — detached HEAD is truncated to an unconditional 7-char prefix (no ambiguity lengthening, no abbrev config), so two colliding commits display identically and a legitimate detached-HEAD move can be suppressed by the change gate. `BranchModel.swift:186,203`
5. **risk** — the benchmark trace is global, not app-scoped: every traced git (including PTY children inheriting the env, or a workload run from a traced shell) counts, and the file accumulates across runs. `bench-git-spawns.sh` (The numbers in the comment above used a fresh trace file per run and an untraced workload shell, so they hold — but the harness should filter by app pid.)

jiweiyuan:
Addressed the review in `bdb1bb8`:

- **Finding 1 (bug)** — the worktree fingerprint now resolves the primary git dir through the `commondir` file and fingerprints each entry's admin `gitdir` contents. Verified live: `git worktree move` previously never changed the name-only fingerprint (0 reconciles); now the next git-dir event reads the moved `gitdir` and fires exactly one `git worktree list`. (Like main's unconditional path, detection still rides the next `.git` event — the watch itself only sees direct children — so parity is restored with correct gating on top.)
- **Finding 2 (bug)** — the build-product event filter checks components *relative to the watched roots*, so a repo living under a `.build`/`node_modules`-named ancestor no longer goes deaf. The tracked-build-products tradeoff stays as documented (heals on the next real event or app activation).
- **Finding 5 (risk)** — the bench harness unsets `GIT_TRACE2_EVENT` in its workload subshell and reports the per-run trace-line delta.
- Findings 3 (shared common-dir double project) and 4 (detached-HEAD 7-char prefix) are accepted as documented tradeoffs — both heal on app activation, and the 7-char chip is a deliberate no-spawn display choice noted in the code.

Re-ran the 45s benchmark on the fixed build: **734** app-spawned gits vs **2,577** on main — the perf win is intact. `swift build` + all 25 tests pass.

PR #104, opened 2026-07-26
Survive cap-truncated UTF-8 and bound grep output by bytes

jiweiyuan:
The original three search-pane commits (redesign, grep hardening, flattened rows) were already absorbed into main through earlier squash merges — after rebasing, this PR's real remaining delta is the codex-review fix only:

- Decode capped grep output lossily: terminating at the line cap can cut the pipe mid-UTF-8-codepoint, and the strict decode dropped every buffered hit (git grep was additionally misread as failed, triggering the recursive-grep fallback).
- Bound the output buffer by bytes (4 MB): newline counting alone let one minified line grow it without limit.

Release Notes:
- Fixed: project search no longer loses results (or falls back to a slow recursive grep) when huge result sets containing multibyte characters are truncated at the cap; a single enormous line can no longer balloon memory during search.

### DIFF
--- a/Sources/termio/FileBrowser/ContentSearch.swift
+++ b/Sources/termio/FileBrowser/ContentSearch.swift
@@ -24,6 +24,10 @@ enum ContentSearch {
     private static let perFileLimit = "20"
     /// Longest line text kept; minified bundles can put megabytes on one line.
     private static let lineCap = 500
+    /// Byte ceiling on buffered tool output. `maxLines` bounds complete lines,
+    /// but the reader only counts newlines — one minified bundle line could
+    /// otherwise grow the buffer without limit before its newline ever arrives.
+    private static let outputByteCap = 4 << 20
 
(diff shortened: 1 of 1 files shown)


jiweiyuan:
Codex review of this PR's diff (2 findings, both fixed in 993b071):

1. **Medium** — terminating grep at the line cap could cut the pipe chunk mid-UTF-8-codepoint; the strict decode then returned nil and dropped every buffered hit (git grep additionally got misread as failed, triggering the recursive-grep fallback). Now decodes lossily (`String(decoding:as: UTF8.self)`).
2. **Medium** — the reader only counted newlines, so a single huge line (minified bundle) could grow the buffer without limit before its newline arrived. Added a 4 MB byte ceiling alongside the line cap.

Codex also verified: subprocess cancellation path is structurally sound, flattened row identities stay globally unique, and row closures preserve the correct URL + 1-based line for editor jumps.

PR #108, opened 2026-07-26
feat(app): native task-completion notifications for agent sessions

### LINKED ISSUE #97: Add native task-completion notifications for Claude Code and Codex sessions
## Summary

Add native macOS notifications when Claude Code, Codex, or another integrated agent finishes a task in a Termio session. This lets users switch to other work and return when the agent is ready.

## Context

A user asked whether task-completion notifications must be configured in the terminal or could be integrated directly into Termio. An app-level implementation would provide one consistent experience across supported agents and shells.

## Desired behavior

- Detect when a supported agent task reaches a completed/settled state.
- Send a native macOS notification with the agent, project, and session name.
- Clicking the notification focuses Termio and opens the originating session.
- Provide settings to enable/disable notifications and notification sounds.
- Avoid duplicate notifications for the same completion event.
- Consider suppressing notifications while the originating session is already focused.

## Acceptance criteria

- [ ] A completed task in a supported Claude Code or Codex session can trigger a native macOS notification.
- [ ] The notification identifies the originating session.
- [ ] Clicking the notification navigates to that session.
- [ ] Users can turn completion notifications on or off.
- [ ] A single task completion does not emit duplicate notifications.

## User request

User request discussing task-completion notifications

jiweiyuan:
Implements #97: a native macOS notification when an agent task settles — finished, or blocked waiting on input — so you can work elsewhere and come back when the agent is ready.

## When a notification fires

The decision chain for a finished turn (`.done`), each gate answering a different question:

1. **Setting on** — General ▸ Notifications, on by default (macOS's own permission prompt is the real gate); sound is opt-in.
2. **Real agent** — a plain terminal's bell/OSC noise never notifies.
3. **termio in background** — the policy Cursor and Codex both ship (`unfocused`-only). While termio is frontmost, the sidebar dot and menu-bar pulse are the completion channel.
4. **Turn ran ≥ 10s** — Ghostty's `notify-on-command-finish-after` idea, tuned for agent turns: quick conversational replies stay quiet. (Warp uses 30s, but only for its unclassified shell-command stream; agent vendors ship 0s — 10s + the surrounding gates fit termio's bimodal turn lengths.)
5. **Turn ran at least one tool** — the task-vs-chat gate: an answer-only turn produced prose the user read as it streamed, not work to come back to. Self-calibrating: applied only to sessions with proven tool telemetry, so an agent without it keeps the duration-only rule instead of falling silent forever.

A blocked agent (`.needsAttention`) skips gates 4–5 — blocked is worse than done (Codex's `approval-requested` / `turn-complete` split): it stalls all progress no matter how fast it got there.

## Tool telemetry (new hook-contract field)

Gate 5 needed a signal the hook contract didn't carry, so this PR extends it, cloning the existing `hooks.conversation` locator pattern: a manifest may declare `hooks.tool` naming the stdin JSON field with the running tool's name (claude.json: `"tool": "tool_name"`), the installer stamps `--tool-from <field>` into hook commands, and the `termio agent report` CLI mines the field into the report. `sync()` refreshes the stable CLI copy before restamping hooks, so the flag and its reader deploy together; existing installs pick it up at next launch. The gate keys off `setCurrentTool` — the tool observations' choke point — so future tool sources feed it without extra wiring.

## Watched vs selected

The status paths used to treat *selected* as *watched* (`selectedSessionID != id` guards predating this feature), which silently swallowed the flagship scenario: watch the session, cmd-tab away, agent blocks — no transition, no notification. A new `TermioStore.isViewing(_:)` (selected **and** `NSApp.isActive`) replaces those guards in the hook-attention, screen, title, and bell paths: a selected-but-backgrounded session now settles to `.done`/`.needsAttention` and notifies, while the selected-and-frontmost case stays quiet as before.

## Mechanics

- **Trigger**: fired from `TermioStore.setStatus`, the single status choke point, on genuine transitions only — one completion → one notification. The working transition starts the turn clock.
- **Content**: agent + verb in the title, project as subtitle, the session's live display title as body, and the agent's brand mark as a `UNNotificationAttachment` thumbnail (macOS reserves the leading icon for the posting app).
- **Click**: the session id rides in `userInfo`; clicking reveals the originating session via `revealSession` — the same verb the `termio sessions focus` CLI and the menu-bar tray now share (select + acknowledge + activate + deminiaturize).
- **Lifecycle**: one identifier per session so a newer settle replaces the banner; per-session state is a single `TurnState` struct (turn clock, tool flags, generation counter) that teardown drops whole; `withdraw` removes by deterministic identifier unconditionally, so banners from a crashed run still clear; engaging from the iPhone (companion attach) withdraws the Mac banner too; `willPresent` returns `.list` so a mid-flight return leaves the notification findable in Notification Center without a banner.
- **Authorization**: requested lazily at the first notification; the post chains on `requestAuthorization`'s own completion, so the very first completion — the one that raises the permission prompt — is still delivered after the user clicks Allow.
- **Safety**: guarded on `AppChannel.isBundledApp`, since `UNUserNotificationCenter` aborts outside a real app bundle (`swift run`, tests).

## Review

Two rounds. A Codex sibling session found 4 issues (in-flight invalidation, attachment-file leak, click-acknowledge, Swift 6 isolation) — fixed in 76d01a4. A second 7-angle review found the dead tool gate, the watched-vs-selected gap, the first-notification permission race, and crashed-run banner staleness — fixed in a4d1803, which also consolidated the per-session state and extracted shared helpers (`NSImage.pngData()`, `AppDelegate.mainWindow`, `AppChannel.isBundledApp`).

References: Ghostty's notification handling (read from source), Warp's notification shim + docs, Cursor's and Codex CLI's notification policies.

Closes #97


jiweiyuan:
Codex review round-trip (sibling session) — 6 findings, 4 fixed in 76d01a4:

- **In-flight race (High)**: fixed — per-session generation counter invalidates a post whose authorization round-trip is still running; the final `add` re-validates generation + setting on the main actor; `withdraw` now clears the pending-request store alongside delivered banners.
- **Temp PNG leak (Med)**: fixed — the attachment copy is created only once the request is certain to be scheduled, and removed if attachment init throws.
- **Click doesn't acknowledge an already-selected session (Med)**: fixed — `revealSession` calls `markSeen` explicitly, since the selection `didSet` skips a same-value write.
- **Swift 6 isolation of `sessionKey` (Low)**: fixed — `nonisolated` statics + `@preconcurrency import UserNotifications`; forced recompile is warning-free.

Declined:
- **needs-input → finished double notification**: intended — those are two distinct events in the status model, and the per-session identifier already replaces rather than stacks banners.
- **Bundle guard "too weak"**: the guard exists to avoid the framework's abort outside an app bundle, not as an auth boundary; the delegate is only installed from termio's own `applicationDidFinishLaunching`.

Issue #105, opened 2026-07-26
SSH sessions: inspector file tree shows the local Mac, not the remote host


jiweiyuan:
## Problem

Open an SSH session (e.g. `ukvps` from New SSH Connection) and the inspector's file tree shows the **local Mac home directory** while the terminal is on the remote box. The panel and the session disagree about which machine they're describing — worse than showing nothing.

The whole FileBrowser stack is local-only by construction: `FileNode` loads via `FileManager.contentsOfDirectory`, watching is FSEvents, content search is local `git grep`.

## Short term (cheap, honest)

Hide the file/search/changes inspector panes for `session.sshHost != nil` sessions (or show a "remote session" placeholder). One conditional; removes the lie without building anything.

## Long term: remote file tree over the existing SSH connection

Recommended route — **reuse the OpenSSH CLI via ControlMaster** (the TRAMP pattern), not an in-process SSH library:

- Terminal sessions already run the user's real `ssh`; opening the connection with `-o ControlMaster=auto -o ControlPath=...` lets `sftp -b` / `ssh host -- ls` piggyback on the authenticated connection — no second handshake, and ssh_config, ProxyJump, agent, sk- keys, 2FA all work for free because OpenSSH does them. Matches the Settings ▸ SSH philosophy (config is the single source of truth; probes already shell out to real `ssh`).
- Production proof: Zed's `crates/remote/src/transport/ssh.rs` (spawns the user's ssh binary as a ControlMaster and multiplexes everything over it); xpipe ships remote browsing over plain shell commands, no SFTP subsystem needed.

Implementation notes:

- Introduce a filesystem-provider abstraction under `FileNode` (local + remote). The shape to imitate is Blink Shell's `BlinkFiles` `Translator` protocol (transport-agnostic walk/stat/read over any backend) — GPL, so reference-only.
- No FSEvents remotely: refresh on demand / app-focus reconcile (same pattern as the git pane), not real-time.
- Content search gets remote almost free: run `git grep` on the host over the same connection.
- Editor write-back is the deep end — start **read-only preview** (same decision as the iOS companion file preview: read-only, size-capped); atomic upload write-back is phase 2.

## Library route (rejected for macOS, noted for iOS)

[...]

PR #115, opened 2026-07-26
Browse an SSH session's remote host in the inspector

jiweiyuan:
Implements #105, **Phase 0 + Phase 1** of the design proposal (P2 search and P3 write-back remain out of scope).

## What changed

**P0 — stop showing local state for a remote session.** When an SSH session is selected, the local file watcher, drop target, git count, and local path resolve away; Search, Changes, and Issues show an honest remote-unavailable state.

**P1 — a read-only remote tree, over OpenSSH's SFTP subsystem.**

- **Reuse the terminal's connection.** The interactive SSH session owns a private, per-process 0700 ControlMaster directory. The pane opens the subsystem over that socket (`ssh -s <host> sftp`) with `BatchMode=yes` and `ControlMaster=no`, checks the socket with `-O check` before launching, and so never redials or raises an authentication prompt.
- **No remote shell.** Listings and reads are SFTP v3 packets rather than `find` and `dd` through the login shell: typed replies, no quoting, and a name carrying a newline, a quote, or a leading dash needs no handling at all. A host whose `~/.bashrc` writes to stdout can no longer corrupt a preview. `RemoteCommand` and `RequestTTY` are pinned off, because either one set in `~/.ssh/config` would stop the subsystem from starting; nothing else about the user's configuration is overridden.
- **One channel, many requests.** Replies are matched by request id, so expanding a folder costs a round trip rather than a process launch, and a preview's chunks are pipelined.
- **Bounded, and fail-closed.** Per-request deadlines, an entry cap per directory, and a 1 MB preview cap refused from the open handle's own size before a byte moves. The type check is `FSTAT` on that handle rather than a second look at the path, so a file swapped after the check cannot be read through. Symlinks, FIFOs, devices, and directories are never previewed.
- **Safe previews.** Remote files land in private lease-owned temp directories under randomized local names. Closing, replacing, cancelling, or losing a presentation race cleans the lease. HTML/HTM/SVG open as read-only source, and a malformed raster can no longer fall through to WebKit.

## Verification

- `swift test` — 243 tests, 0 failures. The SFTP client is exercised against `/usr/libexec/sftp-server`, the same binary `sshd` invokes, so the protocol is covered end to end with no network and nothing mocked: hostile filenames, every entry kind, multi-chunk reads across pipelined rounds, the size cap, refusal of non-regular files, and 16 concurrent reads on one channel.
- `swift build -c release`, `scripts/check-strings.sh`, and `git diff --check` — clean.
- A live-host pass is still worth doing for host-specific SSH configuration: connect through a config alias, browse and preview, then close the terminal and confirm the pane reports the connection is gone rather than reconnecting on its own.

Release Notes:

- Browse an SSH session's host from the inspector, over the connection the terminal already opened. Files open in a read-only preview; the pane says so plainly when the connection closes.

Refs #105


jiweiyuan:
Reviewed and verified against a real host (ubuntu/ukvps):

- `ssh <host> -- <cmd>` — confirmed modern OpenSSH handles the `--` after the destination (exit 0), so the run path is correct as written.
- `LC_ALL=C ls -lAn --full-time` output matches the parser's 8-metadata-field assumption, and Swift's `split(maxSplits:omittingEmptySubsequences:)` was verified to keep padded numeric columns out of the name field and preserve names with internal spaces.
- `swift build` passes.

One real bug found and fixed in 11fca65: the ControlPath lived under Application Support, where the dev channel's socket path plus ssh's 40-hex `%C` hash measures 104 bytes — over the 103-byte `sun_path` cap, so ssh would refuse every dev-channel session with "ControlPath too long" (release had 3 bytes of headroom before a longer username broke it too). Sockets now live at `~/.termio[-dev]/ssh-mux` (77 bytes).

Issue #129, opened 2026-07-27
session-control.sock: accepted fds leak into spawned shells (no FD_CLOEXEC); read-only list forced onto @MainActor


jiweiyuan:
Two related problems in the session-control server that together turn a busy app into a fully wedged one.

## 1. Accepted connection fds leak into every spawned process

On a long-running app, `lsof -U | grep termio/session-control.sock` shows ~130 lingering connections held by unrelated `zsh` / `claude` / `node` children. Cause: the accepted socket descriptor is **not marked close-on-exec**, so every shell/agent the app spawns inherits the live fd and pins the connection open. The roster of half-open connections grows without bound over uptime and amplifies main-thread churn (see #128).

Fix: set `FD_CLOEXEC` on the accepted descriptor (and on the listening socket) right after `accept()`.

`Sources/termio/Agents/SessionControl.swift` (accept path around `handle(_ descriptor:)` `:146`).

## 2. Read-only `list` is forced onto the main actor

The request callback is typed `@MainActor` (`SessionControl.swift:65`) and every request runs via `Task { @MainActor in … }` (`:174`/`:195`). A read-only `list` therefore contends with SwiftUI render work on the main thread; when the main thread is saturated the CLI hits its 15s timeout even though nothing is actually blocked.

Fix direction: let read-only verbs (`list`, `read`) snapshot the roster off-main (or from an actor-isolated cache) instead of requiring a main-thread slot, so status queries stay responsive under render load.

## Symptom

```
$ termio sessions list
termio: no reply from the app within 15s — it may be busy or wedged
```

Reinstalling the CLI does not help — this is entirely app-side. Verified on a 0.21.1 (583) instance up ~14h. Pairs with #128 (the render-side saturation that these two problems amplify).

jiweiyuan:
Superseded: the `sessions list` timeout was a git-pane refresh storm → #134, not these. fd-leak is harmless; `list` off-`@MainActor` is optional insurance — can reopen focused if wanted.

PR #131, opened 2026-07-27
perf(sidebar): drive the working comet at 15fps instead of display rate

### LINKED ISSUE #128: Sidebar comet uses unthrottled TimelineView(.animation) → main-thread starvation, sessions CLI times out
Observed as: `termio sessions list` (and every `termio sessions …` verb) printing

```
termio: no reply from the app within 15s — it may be busy or wedged
```

on a long-running app, while the app is otherwise up and showing ~20%+ CPU at idle. Reinstalling the CLI does not help — the CLI is fine; the **app main thread is saturated**.

## Root cause

Every session-control request is handled on `@MainActor` (`Sources/termio/Agents/SessionControl.swift:65`, dispatched via `Task { @MainActor in … }` at `:174`/`:195`). So even a read-only `list` must get a slot on the main thread. When the main thread is pegged, the request queues behind render work and blows past the CLI’s 15s deadline.

What pegs the main thread is the working-state comet indicator:

`Shared/Sources/TermioShared/SessionStatus.swift:89`
```swift
TimelineView(.animation) { context in   // .animation ticks every display frame
    grid(phase: p)                       // rebuilds the 3×3 grid + AgentIconView each frame
}
```

`.animation` ticks at the display refresh rate (60/120fps) **as long as the view is on screen, whether or not anything changed**. Each working session row owns one such TimelineView, so cost scales linearly with the number of working rows rendered (sidebar + tray roster). Sampling a wedged instance shows the main thread hot in exactly this path:

```
TimelineView<>.init → WorkingIndicator.body → WorkingIndicator.grid(phase:)
  → _copySequenceToContiguousArray → AgentIconView
```

Over long uptime with a large roster the main thread never catches up, and the session-control server (also on main) starves.

## Fix direction

- Stop self-animating via unthrottled `TimelineView(.animation)`. `MenuBarController` already models the right approach: one shared `1/15s` timer advances the comet phase for all working rows at once (`Sources/termio/App/MenuBarController.swift:303-311`). The sidebar should converge on the same shared low-frame-rate tick instead of N independent 60fps timelines — and stop ticking off-screen.
- Secondary: `MenuBarController.refresh()` rebuilds the whole `NSMenu` on `store.objectWillChange` (`:46-48,61-67`), also on main — worth throttling/diffing.

## Related

Pairs with the session-control fd-leak + main-actor issue (filed separately). Verified against a running 0.21.1 (583) build; the `TimelineView(.animation)` path still exists on `main`.

jiweiyuan:
Fixes #128.

## Problem

`WorkingIndicator`'s self-animating path used `TimelineView(.animation)`, which ticks at the **display refresh rate** (120fps on ProMotion), and every working session row runs its own. With a large roster the main thread saturated re-rendering the spinners — which starved the session-control server (also on `@MainActor`) and made `termio sessions list` time out with `no reply from the app within 15s`.

## Fix

The 1.1s rotation reads identically at a fixed **15fps** (the rate the menu-bar comet already uses), so switch the schedule to `.periodic(from:by: 1/15)`. One-line behavioural change; the wall-clock phase math is untouched.

## Measured

A 30-row harness rendering the same spinner three ways:

| variant | main-proc CPU% | grid rebuilds/sec |
|---|---|---|
| `.animation` (display rate) | ~30 | ~2400 |
| **`.periodic` 15fps (this PR)** | **~12** | **~300** |
| shared one-clock-for-the-sidebar (prototype) | ~15 | ~300 |

**~8× less main-thread work**, no visible change to the animation.

### Why not a shared clock?

A single `CometClock` broadcasting one phase through the environment was prototyped and benchmarked. It gave **no further win**: the number of *visible* working rows is bounded by screen height + `List` virtualization, so the per-row timers never actually scale, and wrapping the whole `List` in one `TimelineView` cost *more* in re-diffing (~15% vs ~12% CPU, and worse at 80 rows). So this PR keeps the minimal per-row change and drops that machinery.

## Scope / notes

- iOS is unaffected — it uses the same self-animating path, now just at 15fps.
- The explicit-`phase` path (the menu-bar comet driving its own timer) is unchanged.
- Companion issue #129 (session-control fd leak + read-only `list` forced onto `@MainActor`) is what lets this degrade over uptime; tracked separately.

### DIFF
--- a/Shared/Sources/TermioShared/SessionStatus.swift
+++ b/Shared/Sources/TermioShared/SessionStatus.swift
@@ -81,12 +81,22 @@ public struct WorkingIndicator: View {
     private let dotSize: CGFloat = 2.5
     private let spacing: CGFloat = 3.6
     private let period: Double = 1.1
+    /// A fixed 15fps cadence rather than `.animation`, which drives the comet at
+    /// the display refresh rate (120fps on ProMotion). A slow 1.1s rotation reads
+    /// identically at 15fps, and it matches the menu-bar comet timer — but a
+    /// sidebar full of working rows then costs ~8× less main-thread work (measured
+    /// 2400→300 grid rebuilds/sec across 30 rows, ~30%→~12% CPU), which is what
+    /// keeps the many-session sidebar from starving the main thread.
+    private static let frameRate: Double = 15
+    /// A stable anchor keeps the phase a pure function of wall-clock, independent
+    /// of when any given indicator first appears.
+    private static let clockAnchor = Date(timeIntervalSinceReferenceDate: 0)
 
     public var body: some View {
         if let phase {
             grid(phase: phase)
         } else {
-            TimelineView(.animation) { context in
+            TimelineView(.periodic(from: Self.clockAnchor, by: 1.0 / Self.frameRate)) { context in
                 let p = context.date.timeIntervalSinceReferenceDate
                     .truncatingRemainder(dividingBy: period) / period
                 grid(phase: p)


jiweiyuan:
Closing: profiling the actually-wedged instance showed the comet spinner is ~2/4000 samples — not the cause of the CLI timeout. The real hog is a GitService `git status/diff` refresh storm on background queues (main thread is idle; `list` times out on main-actor contention, not rendering). Pivoting to the real fix (get read-only `list` off the main actor + throttle the git refresh). This 15fps micro-opt isn't worth carrying on its own.

PR #138, opened 2026-07-27
Add highlight annotations and find-in-document to the PDF reader

brelian:
## Summary

- The PDF reader (opened via **Open in Right Panel** or the full-window overlay) now sports a Preview.app-style toolbar with a highlight split-button, a color/style popover, and a find-in-document field.
- User-created highlights, underlines, and strikethroughs persist to a per-document sidecar under Application Support, keyed by the PDF's content hash. The original PDF is never modified.
- Search runs \`PDFDocument.findString\` with case-insensitive matching, highlights every hit, and offers Return-to-next / prev / next navigation with an \"N of M\" counter.

image

## Design

- **Sidecar over embedded annotations.** Sidecar JSON at \`~/Library/Application Support/termio[-dev]/pdf-annotations/<sha256>.json\`, keyed by the PDF's content hash so moves/renames keep annotations attached. Read misses are silently \"no annotations\", not an error. Every add/delete triggers a full re-serialize + atomic write.
- **Two interaction paths matching Preview.**
  - Interaction A (armed): click the pencil with nothing selected → the tool enters an armed state (visible fill in the current color). Any subsequent completed text selection is auto-annotated. Click again to disarm.
  - Interaction B (one-shot): drag to select text, then click the pencil or pick a style. That one selection is annotated; armed state is not entered.
- **Custom \`.popover\`, not \`Menu\`.** SwiftUI \`Menu\` coerces label icons to template symbols, dropping swatch colors. The popover renders full-color swatches, a Preview-style checkmark rail, plus Underline and Strikethrough rows.
- **Mouse-up-driven armed annotation.** \`.PDFViewSelectionChanged\` fires continuously during a drag, so armed annotation listens on a local \`NSEvent\` monitor for \`.leftMouseUp\` instead. That gives exactly one annotation per drag.
- **Selection cache.** Toolbar buttons steal first-responder focus and can clear \`PDFView.currentSelection\` between the click and the coordinator callback. The coordinator caches the last non-empty selection so the mark actions fall back to it.
- **PDFView subclass for context menu.** Overrides \`menu(for:)\` so we can add \"Delete Annotation\" for termio-owned annotations without stripping PDFKit's own Copy/Look Up/Zoom items.
- **Initial fit.** \`autoScales\` leaves the scale at 1.0 until first layout, so we retry \`scaleFactorForSizeToFit\` for a few runloop ticks until it returns a non-zero factor, then also set \`minScaleFactor\` and \`maxScaleFactor\`.

## Non-goals

- No zoom controls, keyboard shortcuts, undo/redo, ink/free-form drawing, sticky notes, or export-annotated-PDF. Those are follow-up work if the reader proves useful.

## Review

Codex reviewed an earlier revision of this branch. Applied fixes:

- \`documentHash\` now cleared at the start of every \`attach\` and only reassigned after a successful hash, so a hash failure on a new URL cannot corrupt the previous sidecar.
- \`updateNSView\` refreshes the hosted toolbar's \`rootView\` on URL change, resetting the query/armed \`@State\` for a new PDF.
- Search field clears highlights on empty query via \`.onChange\`, not only on Return.
- Armed annotation is driven by a \`.leftMouseUp\` local event monitor instead of a debounce over \`.PDFViewSelectionChanged\`.
- The \`.PDFViewSelectionChanged\` observer is re-registered on every \`attach\`, so a coordinator paired with a new \`PDFView\` still receives notifications.
- \`hasSelection\` and \`applyOnce\` fall back to a cached last-non-empty \`PDFSelection\` when the live one has been cleared by focus loss.
- \`pdfView.menu\` is no longer replaced; a \`PDFView\` subclass overrides \`menu(for:)\` and appends the delete item only for termio-owned annotations.
- Toolbar background switched to \`NSColor.windowBackgroundColor\` + a separator hairline; the \`.thinMaterial\` was illegible in the dark termio theme.
- \`NSColor.hexString\` gracefully handles non-sRGB color spaces instead of trapping on \`redComponent\`.
- \`scaleFactorForSizeToFit\` retries up to 8 runloop ticks in case the initial hop still runs before final layout.

## Test plan

- [x] Open a PDF in the right panel — toolbar renders inline, filename left-aligned, controls right-aligned.
- [x] Click pencil with nothing selected → armed (color fill behind the pencil). Drag to select → auto-highlighted on release. Click pencil → disarms.
- [x] Drag to select text → click pencil → that selection is highlighted, armed state stays off.
- [x] Popover shows five color rows (Yellow / Green / Blue / Pink / Purple) with real color swatches, plus Underline and Strikethrough.
- [x] Right-click a termio highlight → \"Delete Annotation\" removes it; the sidecar updates immediately.
- [x] Right-click a link inside the PDF → PDFKit's own menu still fires (Copy, Look Up, etc.).
- [x] Type a search query → Return jumps to the first match with visible yellow highlight; up/down chevrons walk matches circularly; Return on the same query advances to the next match; clearing the field clears all highlights.
- [x] Open a different PDF in the same panel — the reader re-fits, the sidecar for the new document loads (or is silent if none), and the search field / armed state reset.
- [x] Annotations survive relaunching the app.
- [x] \`swift build\` clean.

## Release Notes

The PDF reader now supports highlight, underline, and strikethrough annotations in five colors (yellow, green, blue, pink, purple), plus a find-in-document field with Return-to-next navigation. Annotations persist per-PDF and never modify the original file.

### DIFF
--- a/Sources/termio/Browser/FilePreview.swift
+++ b/Sources/termio/Browser/FilePreview.swift
@@ -1,5 +1,4 @@
 import AppKit
-import PDFKit
 import SwiftUI
 import WebKit
 
@@ -26,12 +25,14 @@ struct FilePreviewView: View {
 
     var body: some View {
         VStack(spacing: 0) {
-            header
-            Divider()
+            // The PDF reader draws its own toolbar with the filename inline (Preview.app style),
+            // so skip the outer header there to avoid two stacked title rows.
+            if kind != .pdf {
+                header
+                Divider()
+            }
             content
         }
-        // Opaque terminal-colored fill so the overlay fully covers the terminal, running up under
-        // the toolbar like the terminal itself.
         .background(Color(nsColor: settings.terminalBackgroundColor).ignoresSafeArea())
         .onExitCommand(perform: onClose)
     }
@@ -59,7 +60,7 @@ struct FilePreviewView: View {
     private var content: some View {
         switch kind {
         case .pdf:
-            PDFPreview(url: url)
+            PDFReaderView(url: url)
         case .web:
             WebPreview(url: url)
         case .image:
@@ -75,25 +76,6 @@ struct FilePreviewView: View {
             } else {
                 WebPreview(url: url)
             }
-        }
-    }
-}
-
-/// A `PDFView` over a file URL, scaled to fit on the terminal background.
-private struct PDFPreview: NSViewRepresentable {
-    let url: URL
-
-    func makeNSView(context: Context) -> PDFView {
-        let view = PDFView()
-        view.document = PDFDocument(url: url)
-        view.autoScales = true
-        view.backgroundColor = .clear
-        return view
-    }
-
-    func updateNSView(_ view: PDFView, context: Context) {
-        if view.document?.documentURL != url {
-            view.document = PDFDocument(url: url)
         }
     }
 }
--- a/Sources/termio/Browser/PDFAnnotationStore.swift
+++ b/Sources/termio/Browser/PDFAnnotationStore.swift
@@ -1,0 +1,124 @@
+import Foundation
+import PDFKit
+import CryptoKit
+
+/// Persists user-added PDF annotations in a per-document sidecar keyed by the file's content
+/// hash — moving or renaming the PDF keeps its annotations attached and the original file is
+/// never modified.
+enum PDFAnnotationStore {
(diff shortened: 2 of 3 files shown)


jiweiyuan:
Closing this — a PDF annotation editor is outside what termio is for right now.

PR #153, opened 2026-07-28
fix(issues): recover from a 403 with reconnect + grant-org-access

jiweiyuan:
## What

The Issues pane could wedge on a **403**. A 401 self-heals (drop the token → back to disconnected), but a 403 — a valid token with **no rights to this repo**, usually an org that hasn't authorized termio — left the pane bound-yet-empty with **no way out**.

## Changes

The error state now offers two exits instead of a dead-end `ContentUnavailableView`:

- **Reconnect** — drops the token and re-runs the device flow, so you can approve under a different account (`IssuesPanelModel.reconnect()`).
- **Grant Org Access…** — opens the OAuth app's connections page, where the org grant a bare reconnect *can't* make actually happens; GitHub re-issues the same token without re-prompting for org grants (`GitHubIssueAuth.settingsURL`, `openConnectionSettings()`).

## Test plan

- Bind a private repo in an org that hasn't authorized termio → 403 → the pane shows the actionable zero state (message in red) with both buttons.
- Reconnect switches account; Grant Org Access opens the connections page.

### DIFF
--- a/Sources/termio/Issues/GitHubIssueAuth.swift
+++ b/Sources/termio/Issues/GitHubIssueAuth.swift
@@ -17,6 +17,13 @@ enum GitHubIssueAuth {
             ?? "Ov23lipSynmaEBOQR1cM"
     }
 
+    /// The user's connections page for this OAuth app — where org access is
+    /// granted. A 403 on a private repo is usually an org that hasn't authorized
+    /// termio, and reconnecting alone won't fix it; the grant happens here.
+    static var settingsURL: URL {
+        URL(string: "https://github.com/settings/connections/applications/\(clientID)")!
+    }
+
     // MARK: Keychain
 
     private static let service = "sh.termio.app.issues"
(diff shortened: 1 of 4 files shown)


jiweiyuan:
## Review verdict: Request changes

- **Medium — all non-401 failures are presented as organization-access failures.** `IssuesPanelModel.swift:143-148` preserves only an error string for every failure other than 401, while `IssuesView.swift:234-249` consequently offers **Reconnect** and **Grant Org Access** for network failures, decoding errors, 404s, 5xx responses, and rate limiting. GitHub can return 403 for rate limits too, where those recovery actions are wrong. Preserve structured error/recovery state and show these actions only for an access-related failure; generic failures need Retry.

Checks that passed:

- A 401 still calls `disconnect()`.
- `reconnect()` deletes the stored token before entering device flow.
- `settingsURL` uses the same environment-overridable `clientID` as device flow and the documented `/settings/connections/applications/:client_id` route.

jiweiyuan:
Addressed in `fcece95`:

- **Non-401 failures no longer all shown as org-access.** A throttling 403/429 (exhausted `x-ratelimit-remaining` or a `retry-after`) is now classified as `APIError.rateLimited`, distinct from a real access 403. The model records a structured `Recovery` — `.reauthorize` **only** for a genuine access 403, `.retry` for rate-limit / 404 / 5xx / network / decode. The error state offers Reconnect + Grant Org Access only for `.reauthorize`, and a plain **Try Again** otherwise.

`swift build` green.

PR #154, opened 2026-07-28
feat(ios): velocity-aware gestures + Reduce Motion fallbacks

jiweiyuan:
## What

Phone motion polish — make gestures interruptible and physical, and degrade cleanly under Reduce Motion.

## Changes

- **Velocity-aware drawer / dismiss** — the spring now takes the pan's release velocity (`setDrawer(…, initialVelocity:)`, fling handoff), so a flick continues into the animation instead of restarting from zero; a partial drag springs back to the nearest rest state.
- **Dismiss easing** — ease-out on screen dismiss and ease-in on the `(+)` menu collapse, so a shrinking card stays solid and leaves no large low-alpha "ghost" trailing behind it.
- **Reduce Motion** — the `(+)` menu and the Home tab-pill selection drop their scale-from-corner / character-move animations for a plain fade when Reduce Motion is on.

(Also carries a re-rendered `AppIcon.png` that was in the working tree.)

## Test plan

- Flick the drawer open/closed → the spring picks up the flick's velocity; a slow half-drag springs back.
- Dismiss a screen → no faint large ghost during the shrink.
- Enable Reduce Motion → `(+)` menu and tab-pill just fade, no scale/character moves.


jiweiyuan:
## Review verdict: Block / request changes

- **High — reversed release velocity is ignored.** `TerminalViewController.swift:928-955` and the interactive-back path use `velocity > 300 || progress > threshold`, then clamp normalized velocity to nonnegative. After crossing the distance threshold, a hard reversal still commits in the old direction and its momentum resets to zero. Give a strong release velocity priority in either direction, use position only for gentle releases, and preserve signed normalized velocity.
- **High — cancelling an interactive back swipe loses keyboard focus.** `TerminalViewController.swift:905-915` resigns first responder on `.began`, but the noncommitted path only restores the screen frame. A short or cancelled swipe returns to the terminal with its keyboard dismissed. Preserve the previous responder state and restore it after the cancel animation settles.
- **Medium — Reduce Motion is incomplete for the home tab pill.** `GlassControls.swift:192-201` suppresses the icon character animation, but the selection capsule still travels between tabs with a 350ms spring. Under Reduce Motion, snap or crossfade the indicator instead of spring-translating its frame.
- **Medium — attachment-menu dismissal still uses ease-in.** `TerminalAccessoryBar.swift:471-483` applies `.curveEaseIn`, which delays the visible response. Use ease-out for the exit, or decouple shrink and alpha timing if the low-alpha “ghost” needs separate treatment. The screen-dismiss change to ease-out is correct.

Checks that passed:

- Normalizing drawer velocity by remaining distance before passing it to `setDrawer(initialVelocity:)` is the correct unit conversion; the defect is the direction decision and clamping around it.
- The attachment menu’s Reduce Motion branch correctly keeps an identity transform and fades.

jiweiyuan:
Addressed in `7aa91e9` (iOS `xcodebuild` green):

- **Reversed release velocity.** A decisive flick now wins over position in *both* directions (drawer open/close pans + interactive-back); the momentum handed to the spring is signed toward the committed target, so a reversal/gentle release settles from rest instead of committing the old way with zeroed momentum.
- **Cancelled back-swipe keyboard.** `.began` records whether the terminal held the keyboard; a non-committed swipe calls `becomeFirstResponder()` to restore it.
- **Reduce Motion tab pill.** The selection capsule now snaps under Reduce Motion instead of spring-translating its frame (the tint still cross-dissolves).
- **(+) menu dismissal.** Decoupled: shrink on `.curveEaseOut` (immediate response), alpha on `.curveEaseIn` (stays opaque till the small end, no large ghost).

PR #161, opened 2026-07-29
feat(inspector): per-session right-side layout (#160)

### LINKED ISSUE #160: Per-session inspector layout (right-side panel state should follow the session, not be global)
## Problem

The entire right-side inspector state is stored as **global singletons** on `TermioStore` — not per-project, and definitely not per-session. Switching terminal tabs force-clears whatever detail was open, so each session cannot keep its own right-side context (one session reading a doc, one with an editor, one on GitHub).

### Current ownership (all global)
- `inspectorTab` (Files/Search/Changes/Issues/Info) — `TermioStore.swift:202`
- `openFileURL` / `openDiff` / `openTrace` / `openIssueDetail` — `TermioStore.swift:114–160`
- `inspectorVisible` / `inspectorMaximized` / `inspectorListCollapsed` — `TermioStore.swift:172–237`

On session switch, `TerminalPane.swift:112–115` sets all four detail overlays to `nil`. The only thing keyed to the session is `inspectorProjectPath` (the root dir the panes read from).

### Related bug (same root)
Maximize a detail (`inspectorMaximized`, global) → switch terminal tab → `refreshDetailPresentation()` tears down the maximize host but the terminal surface underneath isn't re-shown → **blank window** (observed in fullscreen). Making layout per-session removes the clear-and-teardown race, so this should disappear.

## Proposed design (minimal, no new deps)

A per-session snapshot replacing the blanket clear:

```swift
struct InspectorState: Codable {
    var tab: InspectorTab = .files
    var detail: Detail? = nil        // .file(url,line,readOnly) / .diff / .trace / .issue
    var visible: Bool
    var maximized = false
    var listCollapsed = false
}
// TermioStore:  var inspectorStates: [Session.ID: InspectorState] = [:]
```

On session switch: **save outgoing session's snapshot → restore incoming session's** (instead of clearing to nil). First visit to a session with no snapshot = today's default (Files / no detail).

Scope: **option 3 — the whole right-side layout is per-session** (detail + tab + inspector open/close + width). Width currently lives in the AppKit `NSSplitViewItem` autosave (`"TermioContentSplit"`), so per-session width means restoring it manually on switch.

## Persistence: state.json, NOT SQLite (VSCode precedent)

VSCode persists exactly this (open editors, layout, active editor) and restores it on restart via **hot exit** — users rely on it. But it stores it in `state.vscdb` (SQLite) scoped **per-workspace**, a scale/perf choice for an Electron app with a heavy extension KV store; it moved off a single `storage.json` for corruption/perf reasons termio is orders of magnitude away from. Also note VSCode scopes to workspace, not to a terminal tab — per-session is finer than VSCode.

termio's `state.json` already fills the same role `state.vscdb` fills. So: make `InspectorState: Codable` and add it to `StateFile.Snapshot`. **No SQLite / CoreData / GRDB** — termio has none and this tiny, rarely-written state doesn't justify introducing one.

Edge case: on restore, validate the file path with `FileManager` before opening; if gone / worktree changed, silently fall back to the session's default (never surface an error).

## Acceptance
- [ ] Each session remembers its own inspector tab + open detail (editor/diff/trace/GitHub) + open/collapsed + width
- [ ] Switching tabs restores that session's layout instead of clearing it
- [ ] Fullscreen + tab-switch no longer blanks the window
- [ ] Layout survives app restart (via state.json), with stale-path fallback
- [ ] No new storage dependency

jiweiyuan:
Closes #160.

## What

The inspector's right-side state — which tab, which open detail (file / diff / trace / PR-issue), and the detail's maximize / list-collapse chrome — was a set of **global singletons** on `TermioStore`, force-cleared on every session switch. It's now **per-session**: each terminal tab keeps its own right-side context.

- New `TermioStore.InspectorState` + `inspectorStates: [Session.ID: InspectorState]` (sits alongside `runtimes` / `surfaces`).
- `selectedSessionID`'s didSet **saves the outgoing session's layout and restores the incoming one's**, replacing the blanket overlay-clear that lived in `TerminalPane`.
- Dead sessions' entries are pruned in `syncRuntimes`.

## Fixes the fullscreen blank-screen bug

Maximizing a detail then switching terminal tabs used to tear the full-window maximize host down to nothing (global `inspectorMaximized` + clear-on-switch racing the surface re-show), leaving a blank window. Restoring the target session's real state instead of clearing removes the race.

## Persistence — state.json, no SQLite

Only the **durable subset** is written to `state.json`: the tab + the open **file** (path / line / read-only). A diff / PR / trace is a live snapshot (siblings, PR status, transcript) that gets re-fetched, so those stay **in-memory only** — matching VS Code's *hot exit* (restore open files, not transient views). On restore the file path is validated with `FileManager`; a deleted file / removed worktree silently falls back to no detail, never an error overlay.

## Scope decision (deliberate)

**Content** is per-session; **width + open/closed** stay global (they belong to the AppKit `NSSplitViewItem` autosave). This matches VS Code / JetBrains, where the sidebar viewlet and the editor area are independent — opening a file doesn't collapse the sidebar. Width-per-session would mean overriding autosave + re-driving the divider on every switch (jank-prone, low value), so it's intentionally left out.

## Performance (100-tab target)

Switch hot-path is **O(1)**: `captureInspectorState` / `applyInspectorState` touch a fixed set of fields, independent of tab count. Memory is O(N) tiny structs (mostly nil fields). The only O(N) work is building the persisted-layout dict inside `persist()`, dwarfed by the pre-existing `projects` encode. No new scaling cliff. (Pre-existing note: `persist()` writes synchronously on the selection-change path — a candidate for debouncing if 100-tab rapid switching needs it, independent of this change.)

## Test
- [x] `swift build` clean
- [ ] File open in session A survives switch away/back
- [ ] Git diff / GitHub issue restore per session
- [ ] Fullscreen + tab switch no longer blanks

### DIFF
--- a/Sources/termio/App/App.swift
+++ b/Sources/termio/App/App.swift
@@ -611,6 +611,24 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
         DispatchQueue.main.asyncAfter(deadline: .now() + 0.15, execute: work)
     }
 
+    func windowDidDeminiaturize(_ notification: Notification) {
+        // Restoring from the Dock can bring the window back under its content minimum, and leaves a
+        // tickless terminal surface (and a mid-load WKWebView in the Issues pane) unpainted — the
+        // "minimize while a GitHub issue loads → tiny + black" report. Clamp a shrunken frame back
+        // up (setFrame ignores contentMinSize), force a relayout, and nudge the surface to repaint.
+        guard let window else { return }
+        let minFrameHeight = window.frameRect(forContentRect:
+            NSRect(origin: .zero, size: window.contentMinSize)).height
+        if window.frame.height < minFrameHeight {
+            var frame = window.frame
+            frame.origin.y -= (minFrameHeight - frame.height)   // grow from the top, keep it fixed
+            frame.size.height = minFrameHeight
+            window.setFrame(frame, display: true)
+        }
+        window.contentView?.layoutSubtreeIfNeeded()
+        store.repaintSelectedSurface()
+    }
+
     /// Lets the right inspector grow with the window: its max width is the golden ratio (0.618)
     /// of the current content width, floored at 420pt so it never shrinks below the old fixed cap
     /// on narrow windows. A static `maximumThickness` capped the inspector at 420pt regardless of
--- a/Sources/termio/App/App.swift
+++ b/Sources/termio/App/App.swift
@@ -998,6 +998,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
         } else {
             maximizedDetailHost?.removeFromSuperview()
             maximizedDetailHost = nil
+            // Removing the host re-exposes the terminal, but a tickless surface won't repaint
+            // itself — nudge it so a switch out of a maximized detail can't leave a blank.
+            store.repaintSelectedSurface()
         }
     }
 
--- a/Sources/termio/Git/GitModels.swift
+++ b/Sources/termio/Git/GitModels.swift
@@ -6,7 +6,7 @@ import SwiftUI
 /// Which inspector pane the trailing column is showing — the file tree, the file
 /// search, the git changes list, the issue tracker, or the session Info pane.
 /// Drives the segmented switch at the top of `FileBrowserView`.
-enum InspectorTab: Hashable, Sendable {
+enum InspectorTab: String, Hashable, Sendable, Codable {
     case files, search, changes, issues, info
 }
 
--- a/Sources/termio/Issues/IssuesView.swift
+++ b/Sources/termio/Issues/IssuesView.swift
@@ -33,7 +33,7 @@ struct IssuesView: View {
         // editor and diff), driven by `store.openIssueDetail`, not pushed in here.
         listPane
             .task(id: repoRoot) {
-                store.issuesModel = model
+                store.registerIssuesModel(model)
                 await model.start()
             }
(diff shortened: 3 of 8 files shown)


jiweiyuan:
## Addressed codex review (5 findings)

**Regressions fixed** (commit `85fa978`):
- **#1 High** — Sidebar tab click cleared `openFileURL` before the selection changed, so the didSet captured an already-emptied layout and the outgoing file was lost. Now only clears when re-tapping the already-selected row.
- **#2 High** — Durable inspector edits (open file / switch tab) don't move `selectedSessionID`, so nothing persisted them. Added a debounced whole-state save funnelled through `refreshDetailPresentation`.
- **#3 High** — Launch restore clobbered a session's seeded layout (programmatic selection fired the didSet, capturing the still-default live inspector). Guarded with `isRestoringInspector`.

**Deeper findings fixed** (commit `df427ff`):
- **#4 High** — `issuesModel` was a single global slot → a restored issue could render against another repo's model. Now a `[repoRoot: IssuesPanelModel]` cache; `issuesModel` resolves to the selected session's repo, list fallback when unloaded.
- **#5 Med** — The fullscreen blank was the tickless surface on a stale frame after the maximized-detail host tears down (not the clear-vs-restore). Nudge the selected surface with a short render pump on teardown (`repaintSelectedSurface`).

Confirmed-sound by the review and left as-is: issue+diff restore order, UUID keying, missing-file fallback, old-JSON decode, O(1) switch hot-path.

Issue #172, opened 2026-07-30
termiod POC: open a remote Linux VPS from Mac (project + first session)


jiweiyuan:
## Parent
Part of #164. Depends on SSH deploy + remote attach working.

## Goal
From the Mac app (or thin CLI first), **open a remote Linux VPS** as a first-class host: pick SSH config host → ensure `termiod` is present → create/attach a session — without the user manually ssh’ing and running tmux.

This is the product answer to “support open new linux vps remotely,” scoped to **existing SSH hosts** first. Cloud “create droplet” is optional stretch.

## In scope (MVP)
- [ ] CLI: `termio remote open <ssh-host> [--path <cwd>] [--agent claude|shell]`
  - ensures deploy (calls setup if binary missing/old)
  - creates session on remote termiod
  - attaches or prints session id for Mac UI later
- [ ] Mac UI (minimal): “Open remote host…” sheet
  - Host = `~/.ssh/config` alias (no credential UI)
  - Remote cwd default `~`
  - Optional agent picker
  - Creates a Project/session row that attaches via SSH→termiod (not bare `ssh -tt`)
- [ ] Sidebar badge: remote host sessions distinguishable from local
- [ ] Honest degraded panes: hide local-only file tree for remote until #105 (or show placeholder)

## Stretch (explicitly optional)
- [ ] “Provision new VPS” via user-supplied cloud token (Hetzner/DO/…) — **not required for MVP**
- [ ] Tailscale discovery of hosts advertising termiod

## Out of scope
- Full remote file browser write-back
- iOS direct-to-VPS without Mac
- Multiplayer sharing

## Acceptance
- [ ] User with working `ssh my-vps` can open a durable remote agent session from termio in ≤3 clicks / one CLI command
- [ ] Closing Mac does not kill the remote agent
- [ ] Reopen Mac → session still in list → reattach works
- [ ] No requirement that the user install tmux on the VPS

## Design refs
- #164 Phase 2 success metrics
- `docs/design/remote-projects.md` (UI shape; backend becomes termiod not ssh+tmux)
- `docs/design/termiod-session-mux.md` Phase 2

jiweiyuan:
**CLI MVP shipped in draft PR #177.**

- [x] `termiod remote open <ssh-host> [--cwd <dir>] [--agent shell|claude|codex] [--name <n>]`: ensures `termiod` is deployed (deploys if missing), **creates a durable remote session**, then attaches. One command.
[...]

Issue #171, opened 2026-07-30
termiod POC: SSH deploy + remote attach on Linux VPS


jiweiyuan:
## Parent
Part of #164. Depends on local Rust host landing in the sibling POC issue (local Unix socket attach).

## Goal
Deploy **`termiod` to a Linux VPS over SSH** and attach remotely with the **same protocol** as local — transport is system OpenSSH only (no custom crypto).

## In scope
- [ ] Cross-compile or build `termiod` for `linux-x86_64` / `linux-aarch64`
- [ ] `termio remote setup <ssh-host>` (or `termiod deploy`) that:
  - SSHs with user’s config/keys
  - installs binary to `~/.local/bin/termiod` (or configurable)
  - starts/restarts user-level service **or** on-demand socket activation / first-attach spawn
- [ ] Remote attach path: `ssh -T host -- termiod attach <id>` (or stdio bridge)
  - Prefer **stdio bridge** over inventing TCP listen on 0.0.0.0
- [ ] Reuse ControlMaster / `RequestTTY` patterns from zmx docs where attach is interactive
- [ ] Session survives SSH disconnect (daemon on VPS owns PTY)
- [ ] Document security: no public port by default; SSH is the ACL

## Protocol
Same v0 as local:
- control + raw PTY
- detach ≠ kill
- list/create/kill over the SSH-tunneled control channel

## Out of scope
- Provisioning a brand-new VPS from a cloud API (next issue)
- Phone direct-to-VPS
- Remote file tree (#105 long-term)
- Grid-diff wire

## Acceptance
- [ ] From Mac: deploy to an existing SSH host in `~/.ssh/config`
- [ ] Start agent/shell session on VPS; disconnect laptop; reconnect and reattach
- [ ] `list` from Mac shows remote sessions
- [ ] No tmux/zellij required on the host for this path

## Design refs
- #164 Phase 2
- `docs/design/termiod-session-mux.md` §4.5 SSH boundary
- `docs/design/remote-projects.md` (interim — this issue supersedes tmux wrap as target)
- zmx SSH workflow (inspiration only)

## Deploy sketch
```
Mac                          Linux VPS
termiod-cli ── ssh ──► termiod (user service)
                  attach/list/create over stdio or ControlMaster mux
```

jiweiyuan:
**Shipped in draft PR #177** (same v0 protocol as local; transport is system SSH only).

[...]

Issue #170, opened 2026-07-30
termiod POC (Rust): local durable PTY host over Unix socket


jiweiyuan:
## Parent
Part of #164 (`termiod` session mux).

## Goal
Ship a **Rust POC** of `termiod` that owns PTYs on the local machine, so detach ≠ kill and a thin client can attach over a Unix socket.

This is the **language spike** for open decision #1 in #164: **Rust for the Linux-first host binary**. Mac GUI integration is out of scope for this issue — CLI client only.

## Why Rust
- Cross-compile cleanly to Linux (deploy target for Phase 2).
- Fits herdr / shpool-class prior art; solid PTY + async ecosystem (`tokio`, `nix`, `portable-pty` / `tokio-pty`).
- Isolation + long-lived process safety for a never-die host.
- Superlogical uses Go for networking — we accept that; our wedge is agent-native later, not matching their stack.

## In scope
- [ ] New git worktree for isolated POC work (e.g. `termiod-poc` branch)
- [ ] Crate layout: `termiod` (daemon) + `termio-session` or CLI subcommand client
- [ ] Multi-session table: create / list / kill
- [ ] Spawn PTY + shell/agent argv; reap on kill
- [ ] Unix socket server (path: `$XDG_RUNTIME_DIR/termiod/…` or uid-scoped tmp)
- [ ] Protocol **v0**: control frames (JSON or length-prefixed text) + **raw PTY binary** stream
- [ ] `attach` / `detach` (detach leaves session alive)
- [ ] Multi-client fan-out of PTY output (read-only second clients OK; single writer default)
- [ ] Resize (`TIOCSWINSZ`) with newest-client claim
- [ ] `send` / `run`-style inject without full interactive attach
- [ ] Smoke tests: attach → type → detach → reattach still has process; quit client, session lives

## Out of scope (this issue)
- libghostty-vt snapshot / scrollback restore (Phase 3 / follow-up)
- SSH deploy / remote VPS (# follow-up child)
- Mac app / iOS Companion wiring
- Agent status (`needs-you`) — host metadata later
- Windows host

## Architecture (steal from zmx, not tmux)
```
client ── unix socket ── termiod
                           ├── session table
                           └── per session: PTY owner + byte fan-out
```
- Hot path = raw bytes over socket (zmx lesson).
- No window/pane manager inside the daemon.
- VT sidecar later; not required for POC accept.

## Acceptance
- [ ] `termiod serve` (or auto-daemonize) stays up after all clients disconnect
[...]

Issue #164, opened 2026-07-30
termiod: agent-native session mux (durable PTY host + multi-client attach)


jiweiyuan:
## Summary

Build **`termiod`** — a multi-session **PTY host / session mux** so every agent or shell lives independent of the Mac window. Mac GUI, iOS Companion, CLI, and remote VPS all become **attach clients** of the same host.

**Design doc (source of truth):** `docs/design/termiod-session-mux.md`  
Parent architecture: `docs/design/session-daemon-architecture.md`

## Why

Today sessions die with the GUI, remote is second-class (`ssh -tt` hacks), and iOS attaches by tapping into **app-owned** PTYs. Superlogical is building a generic mux foundation; herdr/cmux/tmux solve slices of this. termio already has the ADE surface (Mac + iOS + agent status + `termio sessions`). The missing piece is the **host boundary**.

**Thesis:** Superlogical builds the highway; we already drive the car. Move the engine out of the dashboard so the car runs on any road.

## POC track (active) — Rust + SSH remote

**Decision (2026-07-30):** spike the host in **Rust**, isolated git worktree, then deploy over **system SSH** to Linux. Do not invent a network stack.

| Issue | Outcome |
| --- | --- |
| **#170** | Local durable PTY host (Unix socket, protocol v0 raw PTY) |
| **#171** | SSH deploy + remote attach on existing Linux VPS |
| **#172** | Open remote Linux VPS from Mac/CLI (project + first session) |

Full plan comment: see latest comment on this issue.

Suggested worktree:
```bash
git worktree add ../termio-termiod-poc -b termiod/rust-poc
```

## Goals (v1)

- [ ] Session lifecycle independent of any viewer (`termiod` owns PTYs)
- [ ] One Session Protocol for Mac / iOS / CLI (promote Companion wire → v0)
- [ ] Local durable host (quit app → agents keep running)
- [ ] Remote host via SSH transport (VPS/devbox agent runtime)
- [ ] Agent status (working / idle / needs-you) produced **on the host**

## Non-goals (v1)

- Generic “multiplexer for all work” platform
- New SSH/mosh/QUIC stack
- Multiplayer product (multi-**viewer** is enough)
- Grid-diff wire on day one (raw PTY v0 is fine)
- Windows host
- Cloud “provision new droplet” on day one (open **existing** SSH hosts first; stretch later)

## Phases

| Phase | Outcome | Tracking |
| --- | --- | --- |
[...]

PR #177, opened 2026-07-30
termiod: durable PTY session host (Rust POC) — #170/#171/#172

jiweiyuan:
> [!NOTE]
> **Updated 2026-08-15 — CI builds termiod now, and the Linux binary has run on a real VPS.**
>
> - **`.github/workflows/termiod.yml`.** No workflow mentioned termiod or cargo before this, so the ghostty pin and the Zig requirement were verified on one laptop and nowhere else. Two jobs, both green on this PR: **macos** installs Zig 0.16.0 by published checksum, points it at the Command Line Tools SDK (Xcode's libSystem is arm64e-only and Zig cannot link against it), then runs `cargo build`, `cargo test`, both smoke suites, and the `aarch64-unknown-linux-musl` release cross build — asserting the result really is a static ARM ELF. **linux** (`ubuntu-24.04-arm`) runs the same suite natively on the platform termiod deploys to. 42 unit + 4 integration tests and both smoke suites pass on each.
> - **`.cargo/config.toml` no longer names a home directory.** Both musl targets linked through an absolute `/Users/…/gcc-ld/ld.lld` path that exists on no other machine, CI included. They now name `rust-lld`, which rustc resolves out of the active toolchain's sysroot, with `linker-flavor=ld.lld` because lld's generic driver will not guess its own entrypoint.
> - **Run natively on `ukvps`** (Ubuntu 24.04, aarch64) against the ghostty `9d8fbd1` engine: the cross-built musl binary passes **84 local + 8 remote** smoke checks there. Note for anyone building *on* that box: Zig's TLS client cannot reach `deps.files.ghostty.org` from it (`ziglang.org` and `github.com` are fine), so a native build there needs a pre-seeded `GHOSTTY_ZIG_SYSTEM_DIR`. GitHub's runners fetch it without trouble.
> - **The flag-off default path is unchanged.** A packaged universal `termio.app` launched without `TERMIO_TERMIOD=1` spawns its session shell as a direct child on a pty — the existing `PTYProcess` path. No `termiod` process starts, the daemon socket's mtime never moves, and the app holds no socket to it.

> [!NOTE]
> **Updated 2026-08-13 — the VT engine is a dependency now, not a vendored tree.** The diff drops from **2,406 files / 814,563 insertions** to **104 files / 19,675 insertions**.
>
> - **`termiod/spike/` deleted** (1,155 files, 397k lines). Both proofs — the alacritty `vt-sidecar` reference and the libghostty-vt `vt-ffi` gate — were standalone binaries outside the workspace, and each carried its own copy of the vendored ghostty tree. They stay reachable at `bc36438`.
> - **`termiod/vt/vendor/` deleted** (1,147 files, 397k lines) along with its hand-rolled `build.rs`. Only `src/terminal`, `unicode`, `simd` and `datastruct` ever reached `-Demit-lib-vt`; the rest was ghostty the app — renderer, font, apprt, config — plus `pkg/` shims for freetype, harfbuzz, glslang and the like. The engine now comes from the `libghostty-vt` crate, which owns the Zig build and the FFI. For reference: herdr vendors the same tree, cmux uses a git submodule plus its own in-repo `ghostty-vt-sys`; neither commits 400k lines twice.
> - **`vt/src/lib.rs`: 910 lines of raw FFI → 443 with zero `unsafe`.** `termiod-vt`'s public surface is unchanged, so `termiod/src/**` needed no edit at all. bindgen and the host `libclang` requirement are gone; Zig 0.15.2 on `PATH` is still required (`libghostty-vt-sys` invokes `zig` by name and honours no `ZIG` override).
> - **`LIBGHOSTTY_VT_SYS_OPTIMIZE=ReleaseFast`** in `termiod/.cargo/config.toml`. The crate otherwise mirrors the Cargo profile, and a Zig `Debug` VT parses too slowly for the sidecar to drain a flood: writes queue ahead of the attach snapshot, `S` never lands, and clients fall back to ring replay. The old `build.rs` forced `ReleaseFast` unconditionally, which is why the vendored build never hit this.
> - **New `termiod/tests/join_invariant.rs`** pins invariant JOIN (§C.5) against a live 200k-line flood — one boundary per attaching client, no gap, no duplicate, no effect on other clients. It is what caught the `Debug`-engine regression above.
> - **Zig 0.16 is not reachable by bumping the pin.** `libghostty-vt-sys` ships pre-generated bindings, so building a different ghostty silently mismatches struct layouts. ghostty `ab0b9da9` corrupts the heap — unrelated `HashMap`s spin forever in the probe loop — and `9d8fbd1`, the ghostty `libghostty-swift` 1.0.16 ships, fails the JOIN invariant plus two snapshot/scrollback smoke checks. Host/client VT parity needs a fork that regenerates bindings against the ghostty it builds, tracked separately.
>
> Verified on macOS-arm64: 42 unit + 2 integration tests, 82 local smoke checks, 8 remote smoke checks, and a static `aarch64-unknown-linux-musl` release ELF.

> [!NOTE]
> **Updated 2026-07-31 — the branch has grown past the v0 POC described below.** Shipped since the original description:
>
> - **Protocol v0.1**: `hello` + capability negotiation, `seq`/`re` request ids, typed error model, `E` event frames, `send`/`wait`, `workstream` in CreateSpec. Live-verified on a real VPS (`ukvps`, aarch64 static-musl): detach ≠ kill across hard disconnect, 12.1 ms median keystroke echo.
> - **#180** — `bytes::Bytes` zero-copy fan-out + bounded 4 MiB per-client backlog (slow clients dropped + logged, RSS bounded).
> - **#179** — non-interactive `attach --observe` (pure copy-to-stdout, never claims the writer); fixes 0-bytes-over-SSH scripting.
> - **Anti-100× benchmark** (`termiod/bench/`): termiod's tee holds 4–6× tmux throughput on the Mac, 16–24× on the VPS; tmux loses ~50% plain→ANSI while termiod barely moves.
> - **#181 v1 VT sidecar** (engine decision: **libghostty-vt**, FFI'd via the new `termiod/vt` crate — same VT as every termio client, fidelity parity):
>   - **Phase 1a**: snapshot-on-attach — per-session VT sidecar thread fed off the hot path; one versioned binary `S` frame (16-byte engine-neutral wire cells) + `ready` event, capability-gated (`snapshot`), exact in-band byte boundary (unit-proven), ring-replay fallback. `fan_out` never blocks on VT parse.
>   - **Phase 1b**: resize is a barrier — quiesce → resize → fresh `S` → resume; `TIOCSWINSZ` failure surfaced as a typed error instead of swallowed; stale in-flight snapshots superseded by `request_id`; promoted writers reclaim size via `ResizeClaim`.
> - Suites: **48 local + 8 remote smoke checks, 10+4 unit tests** — green on macOS-arm64 and native aarch64-musl (ukvps).
>
> **Remote UI landed** (`55b2734`+`8e86866`): per-project `+` ▸ **New Remote Terminal** picks an `~/.ssh/config` host and opens a shell running there (auto-deploys termiod if missing); project right-click **Clone on Remote…** clones the repo onto the host via `ssh … git clone` and opens a terminal in it. Per-session remote host/cwd, all gated behind the opt-in flag — local path byte-identical. This is the remote feature made usable from the app, not just the CLI.
>
> **Remote sessions in the app now work over SSH** (`b91d180`): `TermiodSessionLink` gained a transport seam — local Unix socket or an `ssh <host> termiod stdio` pipe. With `TERMIO_TERMIOD_REMOTE=<host>`, an agent runs on a Linux box and the Mac attaches over SSH, reusing the whole snapshot-repaint client unchanged. Verified end-to-end against a real VPS (hello → attached → S snapshot → ready over `ssh`). This is the differentiated feature — your agent lives on the server, the window is just a view.
>
> **`termiod stdio` bridge landed** (`cbe9e23`): the framed protocol now crosses SSH byte-identical (`ssh <host> termiod stdio`), proven by a deterministic Rust integration test. This is the enabler for remote sessions *in the app* — the differentiated feature — reusing the Mac client below over an SSH pipe instead of the local socket.
>
> **Mac app is now an attach client (#170, in progress).** `TERMIO_TERMIOD=1` runs terminal sessions inside the local termiod daemon; quitting termio detaches instead of killing, and relaunch reattaches by session UUID (same pid) with a clean `S`-snapshot repaint of the current screen — the "quit the app, your agent keeps working" demo. New Swift client in `Sources/termio/Terminal/Termiod/`. Verified at the code layer (build + decode/render against a real captured daemon S frame); the live GUI recording is pending.
>
> Continuity doc for orchestration: `termiod/HANDOFF.md`. Landed since: **Phase 1c** (`attached` carries authoritative rows/cols) and **Phase 1d** (staged scrollback — `H` frames, newest-first, capture-at-boundary via libghostty-vt history grid refs, 1 MiB cap, `snapshot`+`scrollback` caps). **Phase 1e** landed too: capability-gated `G` dirty-row diffs (v1.1 plane) — `grid_diff` clients get S keyframes + dirty-row `G` frames instead of raw `D` (zero D backlog), single ordered sidecar channel prevents S/G reordering, periodic keyframes, raw clients byte-for-byte unaffected. #181's protocol ladder (S → H → G) is complete; QUIC stays gated on measured need.

Rust POC of **termiod**, the durable PTY session host from the epic #164. One binary is both daemon and client; sessions live in the daemon and outlive every viewer. **Detach ≠ kill.** Everything lands under `termiod/` at the repo root so it doesn't touch the SwiftPM app.

Scope follows the design docs' v0: **raw PTY bytes over a socket**, no VT/grid snapshot, no custom crypto (remote = system SSH stdio). The zmx lesson — session persistence only, no in-daemon window manager.

### #170 — local durable host ✅ (acceptance fully met, 16/16 smoke checks)
- protocol v0: framed `[kind][len][payload]` — control JSON + raw PTY data + resize
- PTY spawned with the **login_tty shape** (setsid + TIOCSCTTY) so agents reflow on resize; async read/write with a **dedicated writer task** that avoids the documented PTY write deadlock
- session actor: multi-client output **fan-out**, **single-writer** input with **newest-client claim**, **TIOCSWINSZ** resize, recent-output **ring replay** on (re)attach
- daemon: session table + Unix socket under `$XDG_RUNTIME_DIR/termiod/` (0600), auto-starts on first client op
- CLI: `serve` / `create` / `list` / `kill` / `send` (inject w/o attach) / `attach` (create-on-missing, Ctrl-\ detaches)
- `smoke_test.py`: attach→type→detach→**survives**→reattach same pid; fan-out; single-writer; resize; inject; kill — **all green**

### #171 — SSH deploy + remote attach ✅ (transport verified; cross-arch install documented)
- **cross-compiles to static musl Linux** (x86_64 + aarch64) with the bundled `rust-lld` — no musl-gcc / zig / Docker (`.cargo/config.toml`)
- `remote deploy` (uname-detect → build/`--bin` → scp → verify), `remote list`, `remote attach` over `ssh -t` stdio
- daemon auto-starts detached (setsid) on the remote, so a session **survives SSH disconnect** for free
- `DEPLOY.md`: cross-compile, optional systemd `--user` unit, security model (Unix socket only, SSH is the ACL)

### #172 — open remote host (CLI MVP) ✅
- `remote open <host> [--cwd --agent --name]`: ensure deployed → create durable remote session → attach, one command, `~/.ssh/config` only
- `remote_smoke_test.py`: 8 checks through a fake-ssh transport shim (list, open, survives disconnect, reattach same pid) — **green**

## Not in this POC (by design)
- libghostty-vt grid snapshot / scrollback restore (later phase)
- Mac app / iOS wiring, host-side agent status
- Real cross-arch scp/install is verified only by the static-ELF build + manual `DEPLOY.md` steps (no Linux VPS available in this session; loopback sshd needs a key I didn't add unprompted)

## Test
```sh
cd termiod && cargo build
python3 smoke_test.py          # #170, 16 checks
python3 remote_smoke_test.py   # #171/#172 orchestration, 8 checks
```

Part of #164. Draft — remote paths want a real Linux VPS pass before merge.

## Release Notes:

Internal only — `termiod` is not yet wired into a shipping build. No user-facing change.


jiweiyuan:
## Architecture framing (Superlogical-clean)

Pushed `docs(termiod): Superlogical-clean host · protocol · clients architecture` so the POC reads as a **session host**, not a CLI product.

**Three parts only** (see `termiod/ARCHITECTURE.md` + `docs/design/termiod-session-mux.md` §0 / §4):

| Part | Role |
| --- | --- |
| **Host** (`termiod serve`) | Owns PTYs / sessions |
| **Protocol** | Transport-agnostic attach contract |
| **Clients** | Mac / iOS / CLI viewers |

**Rules:** local = remote to localhost · SSH is a pipe · PTY only on the host · no nested WM · no invented crypto.

Mirrors Superlogical’s *durable session as the missing layer* sequence, with termio’s narrower agent-native wedge.

PR #189, opened 2026-08-01
fix(menus): strip preview right-click menus to their real actions

jiweiyuan:
## Summary
Continues the context-menu cleanup (c5e39e6, #187) across the read-only preview surfaces — each now offers exactly the actions that fit it, instead of the OS-injected grab-bag:

- **Image preview** (`FilePreviewView`): adds a context menu with **Copy Image** — the `NSImage` goes on the pasteboard as bitmap data, so it pastes into chat/design apps as a picture, not a file path — and **Reveal in Finder**.
- **Markdown reader**: keeps the menu stripped to Copy, and adds **Reveal in Finder** beside the existing Close (the reader now knows its on-disk `fileURL`).
- **Diff overlay** (`DiffTextView`): overrides `menu(for:)` to return a minimal **Copy + Close**, dropping NSTextView's read-only defaults (Look Up / Translate / Speech / Share / Services), matching the editor's and reader's stripped menus.

## Testing
- `swift build` clean with all three changes applied.

### DIFF
--- a/Sources/termio/Browser/FilePreview.swift
+++ b/Sources/termio/Browser/FilePreview.swift
@@ -74,6 +74,18 @@ struct FilePreviewView: View {
                     .scaledToFit()
                     .padding(24)
                     .frame(maxWidth: .infinity, maxHeight: .infinity)
+                    .contentShape(Rectangle())
+                    .contextMenu {
+                        Button("Copy Image") {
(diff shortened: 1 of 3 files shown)


jiweiyuan:
Superseded: this content landed on main via #184 (MarkdownReaderView, DiffTextView) and #185 (FilePreview Copy Image) — all three files match main byte-for-byte.

PR #190, opened 2026-08-01
fix(inspector): keep the file list mounted beneath the detail overlay

jiweiyuan:
## Summary
Fixes the file tree collapsing back to its root state after closing a file opened from the inspector.

`List(children:)` holds the tree's disclosure state inside the view, and both of the old hidden states destroyed it: the narrow detail-only branch unmounted the list entirely, and squeezing it to zero width tears down its backing `NSOutlineView` just the same. The fix keeps the list **always mounted at real size**:

- The list stays laid out full-width in every state; when the detail covers it (narrow / list-collapsed fallbacks) it turns inert — `allowsHitTesting(false)` + `accessibilityHidden` — beneath the opaque `InspectorDetailContent`.
- The list's frames form a single structural branch (no if/else between fixed-width and flexible), so it keeps one SwiftUI identity across the two-column ⇄ full transition and never resets.
- The drag divider is layered above the detail, which starts at `width + 1` and would otherwise swallow the trailing half of the divider's 10pt grab strip.

## Testing
- `swift build` clean.
- Manual recipe: expand nested folders → open a file → close it → the expansion state survives, in both two-column and narrow layouts.

### DIFF
--- a/Sources/termio/FileBrowser/InspectorDetailHost.swift
+++ b/Sources/termio/FileBrowser/InspectorDetailHost.swift
@@ -36,26 +36,48 @@ struct InspectorRoot: View {
             let showDetail = store.isDetailPresented && !store.inspectorMaximized
             // Two-column when there's room AND the user hasn't collapsed the list to focus the detail.
             let twoColumn = showDetail && geo.size.width >= Self.twoColumnMinWidth && !store.inspectorListCollapsed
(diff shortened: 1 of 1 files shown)


jiweiyuan:
Superseded: the identical fix landed on main via #185 — this branch's InspectorDetailHost.swift matches main byte-for-byte.

PR #205, opened 2026-08-06
Rework the diff's washes, bands, and intraline spans

jiweiyuan:
Three changes to how the diff pane reads, taken from github.com's review surface and Pierre's [`@pierre/diffs`](https://diffs.com).

### Washes are mixed, not composited

`NSColor.systemGreen.withAlphaComponent(0.13)` over `settings.terminalBackgroundColor` drifts in hue and strength with whatever background is configured, so the add and delete sides stop being equally legible from theme to theme. `DiffPalette` mixes a fixed fraction of the base color *into* the real background in Oklab instead, with bases picked per appearance and a stronger mix in dark mode. CSS's `color-mix(in lab, …)` is the same idea; Pierre uses `lab` only because Cursor and VS Code ship a Chrome with an `oklch` bug, which does not apply to AppKit.

A changed row's gutter now takes a stronger mix than its body — github.com anchors its number cell the same way — and the digits stay in the shared muted ink.

### Collapsed runs get reveal buttons

The band's gutter becomes a filled cell holding one button per available direction: an arrow over a dotted line, the dots standing for the hidden lines and the arrow for the direction they come from. Each reveals 20 lines from that end; clicking the row itself still expands the whole gap. A run with no code above it offers no downward button, and likewise upward.

The band also names the range it hides — `128–264`, plus git's section heading where the run came from a hunk boundary — instead of counting itself. "137 unchanged lines" describes the widget; a line range describes the code.

Two deliberate departures from github.com: the band stays neutral grey rather than blue, and the buttons sit side by side in a single-height row. github.com's double-height expander exists to fit two 16pt icons; a row here is barely taller than its text.

### Intraline emphasis is a word diff

`emphasisRanges` peeled a common prefix and suffix off a modified line pair, so `foo(alpha, beta)` → `bar(alpha, gamma)` had no common middle to strip and had to emphasize everything between the first and last change. `DiffIntraline` word-diffs the pair and marks both edits, leaving `(alpha, ` alone. Spans merge across unchanged whitespace, never open or close on whitespace unless whitespace *is* the change, and `DiffRow.emphasis` widens from `Range<Int>?` to `[Range<Int>]`.

### Tests

24 new tests across three files, all pure logic: the Oklab round trip and the mix's endpoints, that a wash stays a visible and evenly matched step from black, white, and both a dark and a light theme background; the fold's ranges, reveal accumulation on a shrinking band, and hunk-heading parsing; and the word diff's two-span, whitespace, CJK, and rewrite-bail cases. `swift test` is at 92 passing.

### Not verified on screen

I could not capture the window myself — this machine's `screencapture` returns `could not create image from display`, so Screen Recording permission is missing for the host. `swift build` and `swift test` pass and the dev app is built and running from this worktree, but the visual result has not been reviewed. Worth a look before merge.

Release Notes:
- Diff add/delete tints now mix into the terminal background instead of layering over it, so both sides stay evenly legible on any theme.
- Collapsed regions in a diff can be revealed 20 lines at a time from either end, and name the line range they hide.
- A modified line with two separate edits now highlights both, instead of everything between them.

### DIFF
--- a/Shared/Sources/TermioShared/DiffIntraline.swift
+++ b/Shared/Sources/TermioShared/DiffIntraline.swift
@@ -10,7 +10,7 @@ import Foundation
 /// edits and leaves `(a, ` alone.
 ///
 /// Offsets are `Character` counts, matching `DiffRow.text`.
-enum DiffIntraline {
+public enum DiffIntraline {
     /// Longer lines are left plain rather than diffed — the pathological cases here are
     /// minified bundles and embedded data, where every span would be noise anyway.
     private static let maximumLineLength = 2000
@@ -18,7 +18,7 @@ enum DiffIntraline {
     /// The changed spans of a deletion/addition line pair, or `nil` when the two share so
     /// little that spans would be noise (a rewritten line reads better as a plain
     /// add/delete pair than as one line-long highlight).
-    static func spans(old oldText: String, new newText: String)
+    public static func spans(old oldText: String, new newText: String)
         -> (old: [Range<Int>], new: [Range<Int>])? {
         guard oldText != newText else { return nil }
         let oldCharacters = Array(oldText), newCharacters = Array(newText)
--- a/Shared/Sources/TermioShared/DiffIntraline.swift
+++ b/Shared/Sources/TermioShared/DiffIntraline.swift
@@ -48,14 +48,15 @@ public enum DiffIntraline {
         let oldSpans = spans(of: oldCore, changed: changed.old)
         let newSpans = spans(of: newCore, changed: changed.new)
 
-        // The same "too different to be worth marking" test the prefix/suffix pass used,
-        // now measured on what the word diff actually kept: at least a fifth of the
-        // shorter line must survive unchanged.
+        // "Too different to be worth marking": at least a fifth of the shorter line must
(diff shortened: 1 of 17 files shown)


jiweiyuan:
Rebased onto main now that #206 has merged, and folded the two diff models into one.

`Shared/DiffModel.swift` shipped as a copy of this branch's *pre*-rework parser, so the clients had already drifted — the phone marked one intraline span where the Mac marked two, and expanded a folded run all-or-nothing where the Mac reveals from either end. The shared parser now owns the word diff (`DiffIntraline` moved into TermioShared), `DiffExpansion`, the band's reveal controls, and the `@@` section heading. `DiffRow` is an alias for `DiffLine`, `GitService` lost its own parse, and `DiffDocument` renders the shared fold instead of folding again.

The phone keeps its own affordance — the whole row is the tap target and asks for `.all`, since there is no room for a per-direction button. Which directions exist is model; where the control is drawn stays per-platform.

`swift build` clean, `swift test` 110 passing, iOS `xcodebuild` BUILD SUCCEEDED. The iOS changes are mechanical (signature adaptation) and compile-verified only — I have not re-run the phone reader on a device or simulator.

PR #206, opened 2026-08-07
Show the Mac's real changes and diff on the phone

jiweiyuan:
The iOS drawer's Changes pane was fake — `numberOfRowsInSection` returned `MockChange.samples.count` unconditionally, and tapping pushed a 62-line hardcoded string rendered as whole-line green/red text in a plain `UITextView`. That was true even on a live companion session. This replaces both ends with real data and a reader built for a phone.

### Wire

Two message pairs: `listChanges`/`changes` and `readDiff`/`diff`, plus `WireChange` and `WireDiff`. The Mac answers from the same `GitService` the desktop git pane reads, so the two lists can't disagree about what "changed" means. Diffs come back with **full file context** so the phone can fold unchanged runs into bands the reader *expands*; past a 2 MB cap the reply degrades to git's default 3-line context and marks itself `fullContext: false`, which turns the bands into fixed markers rather than lying about what was sent.

Nothing under `Sources/termio/Git/` is touched — #205 owns those files. The server calls into them and reassembles the parsed rows into unified-diff text for the wire.

### Shared diff model

`DiffParser` in `TermioShared` parses unified-diff text into numbered lines, marks intraline spans, and folds long unchanged runs — the desktop's `GitService`/`DiffDocument` logic as pure Foundation. The phone renders from it today; the desktop can adopt it once #205 lands, instead of the two ends drifting.

### The reader

Phone constraints, not a shrunk desktop pane:

- **No horizontal scroll.** Lines soft-wrap, and the hanging indent is computed from the line's *own* leading whitespace, so a wrapped comment stays visually inside its line instead of jumping back to column 0.
- **A saturated 3pt leading bar with a very light wash**, not a full-line fill. The file's own doc comment already argued for this and it was never implemented; at phone size a full-strength green makes syntax colors unreadable.
- **Syntax highlighting** through the same Highlightr pipeline as `FileViewerController`, each diff side highlighted as one text so block comments and raw strings keep their real state.
- **One line-number column** (34pt): a deletion shows its old number, everything else the new one. Two columns would cost a tenth of the width.
- **Fold bands are full-width 44pt rows.** The Mac's gutter-cell buttons don't transfer to a thumb.
- **A "3 of 12" walker** between changed files and a floating jump-to-next-hunk button, so reading a review doesn't mean bouncing back to the list.
- **Long-press → Send to Agent** bracketed-pastes the selected code into the session's prompt and steps the drawer aside. Band labels are stripped from the selection, so the agent gets code, not "⋯ 24 unchanged lines". This is the thing a phone diff can do that a Mac diff doesn't need.

### Files

The tree drilled 18pt deeper per level on a 390pt screen. It now pushes one directory per screen — Apple Files' shape — with the folder name as the title, the parent path as the navigation prompt, and the back button walking out. Repo-wide filename search is unchanged.

Changes is also now the **default** pane: on a phone the question is what the agent just touched.

### Verification

- `swift build`, and `swift test` at 81 passing (28 new: fold ranges and reveal, hunk-gap bands, line numbering, intraline spans, wire round-trips including awkward paths and an older peer's missing flags, and the row→text→`DiffParser` round trip).
- **Mac handlers, live** against a dev build over the companion socket: 39 changed files with correct status letters, ±counts and binary flags; a full-context diff (506 lines carrying an 8-line change); a binary file answering `binary: true` with no text; an untracked file diffing as `@@ -0,0 +1,128 @@`.
- **Phone, live on the simulator** against real git data: the Changes list, a diff rendering with syntax colors and intraline emphasis, tapping a band revealing its 79 hidden lines, the walker moving 1 of 12 → 2 of 12, the changes-row context menu, Files drilling into `.github`, and "Send to Agent" putting `\x1b[200~…\x1b[201~` on the wire.

Release Notes:
- The iPhone app's Changes tab now lists the Mac's real working-tree changes, and tapping one opens a full diff with syntax highlighting instead of a sample.
- A diff on the phone soft-wraps instead of scrolling sideways, marks added and removed lines with a colored edge bar, folds long unchanged runs into a tappable band, and walks between changed files without going back to the list.
- Selecting code in a diff offers "Send to Agent", which pastes it straight into that session's prompt.
- The Files tab opens one folder per screen instead of indenting deeper with every level.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -80,6 +80,9 @@ let package = Package(
             name: "termioTests",
             dependencies: [
                 "termio",
+                // The shared diff model the iOS reader parses with — pure logic,
+                // so it is tested here rather than in an Xcode-only test bundle.
+                .product(name: "TermioShared", package: "Shared"),
             ],
             path: "Tests/termioTests",
             swiftSettings: [
--- a/Shared/Sources/TermioShared/DiffModel.swift
+++ b/Shared/Sources/TermioShared/DiffModel.swift
@@ -55,13 +55,12 @@ public enum DiffParser {
         var id = 0
         var oldNo = 0
         var newNo = 0
-        for raw in text.split(separator: "\n", omittingEmptySubsequences: false) {
-            let line = String(raw)
+        for line in text.split(separator: "\n", omittingEmptySubsequences: false) {
             if line.hasPrefix("@@") {
                 if let (o, n) = parseHunkHeader(line) { oldNo = o; newNo = n }
                 // Hunk rows carry their start numbers so a renderer can size the gap to
                 // the previous hunk when it draws the boundary as a band.
-                rows.append(DiffLine(id: id, kind: .hunk, text: line, oldLine: oldNo, newLine: newNo))
+                rows.append(DiffLine(id: id, kind: .hunk, text: String(line), oldLine: oldNo, newLine: newNo))
                 id += 1
                 continue
             }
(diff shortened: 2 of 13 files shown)


jiweiyuan:
Follow-ups deliberately left out, both blocked on files #205 owns:

- `DiffParser` in `TermioShared` still duplicates `GitService.parseDiff` / `DiffDocument.displayItems`. Once #205 lands, the desktop should adopt the shared model and delete its copies — the band's count-vs-range difference is a render-time choice, since `DiffItem.band` already carries the range.
- `handleReadDiff` reassembles unified text from parsed rows because `GitService.loadDiffText(_:_:context:)` is private; exposing it (or adding `context:` to `diffText(for:in:)`) removes a parse, a 2 MB string join, and the `unifiedText` helper with its tests.
- The companion's `.error` carries no request id, so the drawer attributes failures by probing its own state in order. An id on the envelope — with correlation moved into `CompanionClient` — would delete that ladder.

PR #222, opened 2026-08-08
Draw the Mac's brand marks through the shared SVG parser

jiweiyuan:
## What

TermioShared has carried the same `SVGPath`/`Scanner` port since the iOS session list adopted it, with the desktop noted as "still carries its own copy until it migrates onto this package". This migrates it: the parser is public in TermioShared and the desktop's 268-line copy is deleted.

## How it was verified

`swift build` clean, 124 tests green. Icon rendering is visual — worth a glance at the sidebar's agent marks in a dev build before merging.

## Release Notes

None — internal deduplication.


jiweiyuan:
Rebuilt the post-merge tree clean; the two parser copies differ only in comment wrapping, so the desktop deletion loses nothing.

PR #224, opened 2026-08-09
Add Simplified Chinese localization with an in-app language picker

jiweiyuan:
## What

- **Settings ▸ General ▸ App language picker**: writes `AppleLanguages` into the app's own defaults domain (the same mechanism as System Settings per-app language) and offers a relaunch; "Later" is the safe default because relaunching ends running sessions. No live switching — AppKit resolves strings at launch, and a hot swap would leave the window half-translated.
- **Simplified Chinese UI**: ~520 strings across the menu bar, sidebar, command palette, all settings tabs, git/issues/session-info panes, file browser, editor, find bar, and terminal context menu, following Apple's zh-Hans conventions (拷贝/访达/编组到…).
- **String Catalog pipeline for a pure-SwiftPM app**: `Localizable.xcstrings` is the single editing source; `scripts/compile-strings.sh` compiles it with `xcstringstool` into checked-in `.lproj` resources, because `swift build` copies `.xcstrings` verbatim instead of compiling it (swiftlang/swift-package-manager#6993). Each `.lproj` ships as a verbatim `.copy` — `.process` lowercases `zh-Hans.lproj`, which CFBundle's case-sensitive language matching then silently resolves to English.
- All lookups go through `localized(_:)`, pinned to `Bundle.termioResources` — `String(localized:)`'s default `Bundle.main` holds no strings in a packaged `.app`.

Brought up to date with `main`: the Interface tab's strings moved with it into Appearance, the new Window menu (Minimize / Zoom / Bring All to Front) and "New Terminal at Home" are localized, the agent-skill and command-line-tool rows follow their rewritten copy, and every string that `main` recapitalized to "Termio" carries its translation over.

## Verification

- `swift build` clean; 206/206 unit tests pass.
- Packaged `termio-dev.app`: both `.lproj` tables present with 520 entries each, and `New Terminal at Home` resolves to 在个人文件夹新建终端 in the packaged bundle.
- English fallback: an explicit `en.lproj` is generated so English never depends on CFBundle's fallback order.

Not in scope: the iOS companion (separate Xcode project) and the session-trace HTML.

Release Notes:
- termio can now speak Simplified Chinese: pick a language under Settings ▸ General ▸ App language.


jiweiyuan:
Codex review applied: system-language resolution for the System item, positional-arg relaunch, explicit dark-slot parameter, build-time catalog recompile, atomic compile-strings output.

PR #240, opened 2026-08-10
Paint the GitHub star count yellow

jiweiyuan:
The GitHub band's whole ask is "A star helps other builders find it," but the star next to the count rendered at the same muted grey as the number, so the one glyph carrying that ask was the quietest thing in the pill.

It now uses a new `--color-brand-gold: #eab308` token — the yellow people already picture on a GitHub star, deep enough to hold its edge on the white pill. The traffic-light `brand-amber-deep` was tried first and read as an orange warning tint; GitHub's own `#e3b341` washed out against white. The count stays grey so only the star carries colour.

Release Notes: The landing page's GitHub star count now shows a yellow star.

### DIFF
--- a/web/landing/src/app/globals.css
+++ b/web/landing/src/app/globals.css
@@ -15,6 +15,9 @@
   --color-brand-red: #ff5252;
   --color-brand-amber: #ffb764;
   --color-brand-amber-deep: #fd9426;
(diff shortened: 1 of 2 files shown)


jiweiyuan:
Vercel preview green — landing-only colour token change, merging.

PR #251, opened 2026-08-10
Ship a universal binary so Intel Macs can run termio

### LINKED ISSUE #248: Ship a universal binary so Intel Macs can run termio
## Problem

The released `.app` is Apple-silicon-only, so Intel Macs cannot run termio. Nothing tells the user that before download — they get the Finder's generic refusal.

## Original requests

> 是不是都抛弃 intel chip 了...

— @coosir, [v2ex/t/1232978#27](https://www.v2ex.com/t/1232978#reply27)

> inter 芯片有计划支持吗

— @LittleZ, [#33](https://www.v2ex.com/t/1232978#reply33)

## Verified state

This is a build configuration, not a dependency blocker:

- The prebuilt terminal core already ships an Intel slice — `GhosttyKit.xcframework` carries a `macos-arm64_x86_64` library with both `arm64` and `x86_64` architectures.
- `scripts/build-app.sh:77` runs a plain `swift build -c "$configuration"` with no `--arch` flags, so it produces a host-architecture binary.
- `.github/workflows/release.yml` runs on `macos-26` (Apple silicon) and calls that script, so every release is arm64-only.
- `TunnelManager.Spec` already branches on `arch(arm64)` vs `amd64` when fetching tunnel binaries, so that path is arch-aware.

## Desired outcome

Releases ship a universal `.app` that launches on an Intel Mac.

## Work

1. Build both slices (`swift build --arch arm64 --arch x86_64`, which emits a universal binary directly) and confirm `build-app.sh`'s bundle/sign/notarize steps carry it through — Sparkle embedding and the Developer ID signature both need to see the fat binary.
2. Verify on real Intel hardware or a Rosetta-free check (`lipo -archs`), and check that the launch path — including the downloaded tunnel binaries and the `termio` CLI copied into Application Support — picks the right slice.
3. Decide whether the DMG stays one universal build or splits per arch. One universal build is simpler and the size cost is a few tens of MB.

## Open questions

- Is there an Intel Mac available to verify on? A universal binary that has never been launched on Intel is a claim, not a feature — the honest fallback is to ship it as untested and say so in the release notes.

jiweiyuan:
The shipped `.app` was Apple-silicon-only by configuration, not by dependency: `build-app.sh` ran a bare `swift build`, so only the host slice was ever bundled, while libghostty's `GhosttyKit.xcframework` has carried `macos-arm64_x86_64` all along.

`build-app.sh` now builds one slice per architecture and lipos them into the bundled executable. SwiftPM's own multi-arch mode (`swift build --arch arm64 --arch x86_64`) is not usable here — it routes the build through the Xcode build system, whose eager-linking step passes `-lghostty` with no matching `-L` and fails with `library not found for -lghostty`. One `--arch` at a time keeps the normal build system. Sparkle.framework (and its XPC helpers) already ship universal from their xcframework, and the SwiftPM resource bundles are arch-independent, so only the executable needs merging.

The dev channel still builds the host slice alone, so the rebuild loop stays at one compile. Before signing, the bundled binary is checked with `lipo -archs` and the build fails if a slice is missing — a dropped slice is otherwise invisible until a user on the other architecture tries to launch it.

`TunnelManager.Spec` resolves `arm64` vs `amd64` with `#if arch(arm64)`, which is evaluated per slice, so each architecture downloads its matching tunnel binary. The DMG stays a single universal build.

Verified on this (Apple silicon) machine against a packaged `.app` from `./scripts/build-app.sh`:

- `lipo -archs termio.app/Contents/MacOS/termio` → `x86_64 arm64`; same for Sparkle, Autoupdate, Updater.app, and both XPC services.
- `codesign -v --arch arm64` and `--arch x86_64` both report valid; `codesign --verify --deep --strict` passes.
- Both slices launch: run through `arch -arm64` and `arch -x86_64`, each reaches the run loop, binds its control sockets, and writes `state.json`. `sample` reports `Code Type: X86-64 (translated)` for the Intel slice.
- The dev channel build still produces an `arm64` bundle.

Not verified: a real Intel Mac. The x86_64 slice has only been exercised under Rosetta on Apple silicon, which does not cover an Intel GPU or Intel-only runtime behavior.

Follow-up, not in this PR: the landing site still says "Apple Silicon" (`web/landing/src/components/sections/faq.tsx`, `layout.tsx`). That copy should change when the first universal DMG ships, not before.

Closes #248

Release Notes:

- termio now ships as a universal binary and runs on Intel Macs as well as Apple silicon.


jiweiyuan:
Added an Intel gate on top: `swift build --arch x86_64` on every PR (free Apple-silicon runner, verified locally — links clean and produces an x86_64 binary), plus a `macos-26-intel` job that runs `swift build` + `swift test` natively after merge. Also merged main in, since it had since added the SwiftLint step to the same workflow.

PR #252, opened 2026-08-10
Keep a hidden split group at its own pane frames

### LINKED ISSUE #245: Hiding a split group resizes its panes, repainting them on return
## Problem

Switching between sessions shows a visible flash while the terminal repaints — but only for users with splits. The always-mounted design already killed this for the plain case; split panes were left out of it.

## Original request

> 切 TAB ，界面重绘时会闪

— @lisi1987, [v2ex/t/1232978#57](https://www.v2ex.com/t/1232978#reply57)

## Mechanism

`TerminalPane`'s header states the fix that already shipped: swapping the visible terminal with `.id(session.id)` made SwiftUI rebuild the surface on every switch, the fresh view laid out from a zero frame, libghostty called `ghostty_surface_set_size`, and the resulting SIGWINCH made the shell repaint its prompt. Keeping every session mounted and toggling only opacity removed the resize, and with it the flicker.

That holds for an ungrouped session. It does not hold for a split group, because the frame a mounted surface gets depends on the *selection*:

```swift
let layout = store.splitRoot?.layout(in: bounds)   // TerminalPane.swift:60
…
let paneFrame = zoomed ? … : layout?.frames[id]
let rect = paneFrame ?? bounds                     // TerminalPane.swift:169
```

`splitRoot` resolves only the group containing the *selected* session (`TermioStore.swift:76`). So when the selection moves to a session in another group — or to an ungrouped session — every pane of the group just left drops out of `layout.frames`, falls to `rect = bounds`, and is re-laid-out at the **full pane size** while invisible. Selecting it again re-lays it out at its half/third frame.

Opacity does not affect layout, so both re-layouts are real. Each one runs the full chain: SwiftUI frame change → `TerminalSurfaceCoordinator.synchronizeMetrics` → `surface.setSize` → the `resize` closure in `TermioStore+TerminalSurface.swift:213` → `PTYProcess.resizeFromHost` → SIGWINCH → the shell or agent repaints. The repaint on the way back in is the flash the user sees.

Net effect: every switch **into or out of** a split group sends two spurious SIGWINCHes to each of its panes and repaints them. A user who never splits never sees it, which is consistent with one report out of a thread this size.

## Second, smaller case

A session that has never been on screen mounts on first selection, and its NSView is laid out from a zero frame — one resize, once per session per app launch. Restoring a saved tree means the first visit to each session pays it.

## Desired outcome

Switching sessions never resizes a surface that is merely being hidden or shown. A pane's grid should change only when its geometry genuinely changes: window resize, inspector toggle, split ratio drag, group/ungroup.

The obvious shape is to keep a hidden group laid out at the frames it had, rather than collapsing it to `bounds` — the layout for every group is derivable, not just the selected one. Whatever the fix, it must not reintroduce a resize on the way back in.

## Verification

Statically traced; not yet reproduced at runtime. A runtime check that does not depend on catching one frame by eye: run `trap 'date +%s.%N >> /tmp/winch.log' WINCH` in a split pane, switch to another session and back, and confirm the log stays empty.

jiweiyuan:
Switching sessions repainted every pane of a split group, which is the flash a user with splits sees on every switch.

`TerminalPane` sized mounted surfaces from `store.splitRoot`, and that resolves only the group holding the *selected* session. Moving the selection elsewhere dropped every pane of the group just left out of `layout.frames`; each fell through to `rect = bounds` and was re-laid-out at full pane size while invisible, then laid out again at its half/third frame when reselected. Opacity does not affect layout, so both were real resizes — SwiftUI frame change → `synchronizeMetrics` → `surface.setSize` → `PTYProcess.resizeFromHost` → SIGWINCH → the shell or agent repaints.

Every group's layout is derivable, not just the selected one, so `SplitNode.paneFrames(of:in:)` lays out all groups in the same rect and `TerminalPane` sizes each mounted surface from that, deciding visibility separately. A group that is off screen keeps the frames it had, so revealing or hiding it changes nothing; only a genuine geometry change — window resize, inspector toggle, ratio drag, group/ungroup, zoom — resizes a surface. The always-mounted design is untouched.

Verification:

- `swift build` and `swift test` pass, with two new `SplitTreeTests` pinning that a pane's frame does not depend on which group is selected.
- Runtime, on the dev build with the issue's probe (`trap … WINCH` in a split pane, switch to a session in another group and back): pre-fix logs **2 SIGWINCHes per round trip**, one each way; with the fix, three round trips log **0**.
- The dev build's view dump confirms the geometry directly: while its group is hidden, a two-pane group stays at its own 253×888 / 252.5×888 frames instead of collapsing to the 506.5×888 pane bounds.

Release Notes:
- Fixed the terminal flashing when switching between sessions while a split group is open. Panes are no longer resized when they are merely hidden or shown, so the shell never repaints on a switch.

Closes #245


jiweiyuan:
Verified at runtime on the dev build: pre-fix, one round trip out of and back into a split group logs 2 SIGWINCHes in a trapped pane (one each way); with the fix, three round trips log 0.

PR #275, opened 2026-08-12
Place the mobile agent GUI on the agent's own config channels

jiweiyuan:
Revises `design/mobile-agent-ui-protocol.md` to v3, answering two questions: how the iOS GUI gets its content for Claude Code and Codex, and how much of per-agent support can live in a manifest instead of Swift.

## Three channels, two usable

Every coding agent exposes structure three ways — lifecycle hooks, an on-disk transcript, and a sidecar session protocol. The first two are zero-intrusion; the third (Codex app-server, Claude `stream-json`, ACP) replaces the runtime, and the TUI is the product. So the content plane comes from the transcript, the interaction plane from hooks, and grid-reading survives only as the observed-level fallback already described in §6.4.

## Where the config line falls

The ten shipped manifests already carry no grid regexes — `titleStatus` reads the OSC title, and the old screen-scrape `status` field is unused. Keeping that property costs two new manifest blocks, `transcript` and `questions`, both carrying locations and key names picked from closed sets, matching how `resume` and `hooks.dialect` already work.

Parsing a transcript shape stays Swift, and the doc says why rather than leaving it to taste: waku spends 300–530 lines of Rust per provider, linkcode 2155 lines on Claude Code alone. Neither configures an agent as pure data, so a JSONPath mapping DSL would buy a mini-interpreter that is hard to write and harder to test.

## Also

- linkcode and waku added to the prior art. Notably waku still reads Claude's JSONL for checkpoints and forking while holding a live protocol stream — outside evidence for "the transcript is the durable truth".
- The open Codex question is narrowed: its approvals are a first-class protocol object on the app-server, which is unreachable while the TUI runs, so the question is whether `hooks.json` has the same write-back.

Docs only — no code paths touched. The `docs/README.md` index was regenerated from front matter.

Release Notes: None

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -74,7 +74,7 @@ from the real front matter).
 | done | research | ["Competitive analysis: Vibe Island family (status monitors)"](competitive-analysis/07-vibe-island.md) |
 | done | research | ["Competitive analysis: Warp (alternative paradigm)"](competitive-analysis/08-warp.md) |
 | done | research | ["Termio differentiation, gaps, and risks"](competitive-analysis/09-differentiation-and-gaps.md) |
-| draft (v2) | design | [移动端 Agent UI 协议 —— PTY 之上的旁路结构面（ACP 词汇）](design/mobile-agent-ui-protocol.md) |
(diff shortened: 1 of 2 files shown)


jiweiyuan:
Closing with #296 — the v3 design it revises is no longer being built.

PR #279, opened 2026-08-12
Restore the Font keybinding category key

jiweiyuan:
Keyboard's three font commands ask for a `Font` category, but the catalog has no such key: renaming the Appearance section header from "Font" to "Terminal font" took it along. It was the only one of the eight keybinding categories missing, and its three commands are all translated — so a Chinese user got an English `Font` heading over 增大字体, 减小字体, 重设字体大小.

Adds the key with 字体, matching `Terminal font` → 终端字体, and regenerates the compiled `.lproj` output.

Release Notes: The Keyboard settings tab now shows its Font category heading in Chinese.

### DIFF
--- a/Sources/termio/Resources/Localizable.xcstrings
+++ b/Sources/termio/Resources/Localizable.xcstrings
@@ -1341,6 +1341,16 @@
         }
       }
     },
+    "Font": {
+      "localizations": {
+        "zh-Hans": {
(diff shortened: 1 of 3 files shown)


jiweiyuan:
Superseded — main already carries `Font` translated as 字体 via #278, and #280 adds the translator comment.

Issue #288, opened 2026-08-12
Add a chat lens to the iOS companion, driven by a structured content plane


jiweiyuan:
The phone renders a session as an 80-column VT grid. That is the wrong surface for an agent conversation: the text is small, approvals are arrow-key menus, and a session whose process has exited shows a blank screen even though its whole conversation is sitting on disk.

This adds a second lens over the *same* session — a native conversation view driven by a structured content plane derived from the agent's own transcript and hooks. The terminal stays exactly as it is, one tap away.

## Why this is not the third attempt at a chat UI

iOS chat UI has been built and reverted twice. Both attempts re-drew what the terminal had already drawn, so they earned nothing. The design (`docs/design/mobile-agent-ui-protocol.md`) sets four capabilities the terminal genuinely cannot provide, and treats them as the bar:

1. A dormant session reads in full — the transcript outlives the process.
2. Diffs render as diffs at phone width.
3. A plan is a checklist, not a redrawn TUI box.
4. A permission question is a button, not a menu you arrow to.

If it cannot do these four, it should not ship.

## Design

- Content plane derived from two zero-intrusion channels: the agent's transcript JSONL and its hooks. Never by parsing the VT grid.
- Flat typed events in ACP vocabulary, not ACP's method layer — the same conclusion Happy reached after trying the ACP shape.
- Tool events are **upserts keyed by call id**, because the source is a file: re-reads, resume and fork all replay the same record, and replay must converge.
- `seq` + `since` cursor, shared with the PTY ring buffer's catch-up so both planes have one reconnect story rather than two.
- One endpoint, separate channels, separate overflow policies: bytes drop and re-snapshot, events never drop.

## Scope

**This PR** — read-only lens, Claude dialect:

- `AgentEvent` in `TermioShared`; `subscribeEvents` / `agentEvents` on the wire.
- `AgentEventStore` + `ClaudeTranscriptNormalizer` on the host, reusing the existing transcript resolution that already works for sessions the Mac never opened.
- `ChatLensViewController` on iOS with native Markdown, tool cards, diffs and plans.

**Follow-ups** — approval cards driven by `PreToolUse` (the interaction plane, independently shippable); Codex dialect; slash-command and `@`-mention affordances; attachments.

## Non-goals

- No headless or ACP runtime. The agent keeps running in a real PTY.
- No regex against any agent's menu layout.
- No daemon in this PR. The plane boundary is the cut line when `termiod` lands.

jiweiyuan:
[...]

PR #296, opened 2026-08-13
Read a session's conversation as a chat lens on the phone

jiweiyuan:
The terminal stays the interface. This adds a second lens onto the same live session, for the times a phone is the only screen you have.

## How it works

The host turns an agent's on-disk transcript into an `AgentEvent` stream that clients subscribe to with a cursor, so a phone reconnecting mid-turn asks for the gap rather than the whole conversation, and a dormant session still reads — the transcript outlives the process. A tool event is upserted rather than re-appended, so a card revises in place instead of jumping to the bottom when it finishes.

The phone renders those events natively — prose, tool cards, diffs, plans — and its composer types into the same PTY the Mac terminal owns. One session, two lenses, no local/remote mode switch and no session restart.

## Sending

A prompt waits for the agent's screen to settle before it is typed. Bytes written while a CLI is still painting are read by whatever holds the PTY at that moment and are simply gone: the message disappears, no turn starts, and nothing streams back — so the lens looked broken in both directions from a single cause. This is a client-side approximation; the Mac is the side that actually knows when the agent is ready, and promoting `prompt` to a typed wire message with an acknowledgement is the follow-up.

## Verified

Driven on device and in the simulator against a real Mac companion and a real Claude session: a prompt sent from the lens into a dormant session reconciles from pending to sent and the reply streams back. `AgentEventStoreStreamTests` pins the half that cold-replay tests miss — that a turn appended *after* a subscriber attaches reaches it.

## Terminal first, chat one tap away

A session opens in the terminal — the product's position rather than a default nobody chose — and each view carries the switch to the other in the same header, so swapping never moves the chrome. Settings ▸ Appearance ▸ Sessions flips which one a session lands on, and back leaves the session from either view rather than making it a two-step.

Reading never starts anything. The Mac creates a session's process on first attach, so the chat subscribes without attaching and only the terminal — or a send — claims the PTY. Asking whether a session is live no longer goes through the call that *spawns* one either, which meant opening a finished conversation used to resurrect it.

## The conversation itself

Events now carry the transcript's own clock, so a conversation read days later groups by the day it happened rather than the day you opened it: day pills between days, a time under your own messages, and grouping that tightens consecutive messages from one speaker. Around that, a jump-to-latest control, long-press to copy, a haptic on send, and a `working…` line under the title fed by the same hook-driven status the session list already broadcasts — the chat was simply not listening to it.

Measured row heights are fed back as estimates. With one fixed guess, a conversation full of forty-line diffs shifts under your finger every time a row turns out taller than the guess, which is the single loudest way a self-sizing table announces itself.

Three dialect details that were each wrong first: a plan is one row that revises in place, folding both shapes Claude Code writes (`TodoWrite` sends a whole list, `TaskCreate` / `TaskUpdate` one task at a time); a revised event keeps its place in the log and only takes a new `seq`, without which a cold replay puts every finished tool card *below* the diff it produced; and the user side of a transcript is filtered of the CLI's own plumbing — `isMeta` rows carrying a skill's body, `<command-name>` envelopes, `<local-command-stdout>` — because rendering those in a user bubble claims the human said them.

## Not in this PR

Approval buttons (the `PreToolUse` interaction plane), the Codex dialect, the `transcript` / `questions` manifest blocks, an on-device event store so a chat opens from disk instead of re-downloading, and a turn-level progress row to collapse the tool traffic of a busy turn.

## Release Notes

- Read an agent session's conversation on your phone, with tool calls, diffs and plans rendered natively rather than scraped off the terminal.

### DIFF
--- a/Shared/Sources/TermioShared/AgentEvent.swift
+++ b/Shared/Sources/TermioShared/AgentEvent.swift
@@ -1,0 +1,236 @@
+import Foundation
+
+/// One event on a session's **content plane** — the structured projection of an
+/// agent's conversation, derived from the transcript the agent writes to disk.
+///
+/// The byte plane (raw PTY frames) and this plane describe the same session and
+/// ride the same connection on different channels. They differ in durability:
+/// bytes die with the process, events live as long as the transcript file, which
+/// is why a dormant session still has a readable conversation and a blank
+/// terminal.
+///
+/// Two properties the wire depends on, both consequences of the source being a
+/// **file** rather than a live stream:
+///
+/// - **`seq` is dense and monotonic per session.** A client that reconnects
+///   asks for everything after the highest `seq` it holds. Same mechanism as
+///   the PTY ring buffer's catch-up, so there is one reconnect story, not one
+///   per plane.
+/// - **Tool events are upserts keyed by `call`.** Re-reading a transcript,
+///   resuming, or forking makes the same tool record appear again; replaying it
+///   must converge rather than duplicate. A start/end event *pair* would not —
+///   which is why this is a single mutable event and not two.
+public struct AgentEvent: Codable, Sendable, Equatable {
+    public enum Role: String, Codable, Sendable, Equatable {
+        case user, agent, system
+    }
+
+    /// ACP's tool vocabulary. Deliberately a small closed set: the client picks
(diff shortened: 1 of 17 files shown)


jiweiyuan:
Closing with #275 and #288 — the chat lens is off the roadmap; the terminal stays the interface.

PR #312, opened 2026-08-16
Compare a branch against the one it would merge into

jiweiyuan:
The git pane showed uncommitted changes and past commits, but not the thing you read right before opening a pull request: everything the branch would carry over. A third tab — **Changes | Compare | History** — answers it.

**Compare** shows the branch measured against the branch it would merge into: a base picker, the changed files with `+`/`−` totals, and the commits the merge would bring. Clicking a file opens its diff over the terminal through the overlay that already existed for the Issues pane's pull-request rows (`GitDiffRequest.range`).

Details that decide whether this is correct:

- **Three-dot range.** `base...HEAD` diffs from the merge base, so the list matches the forge's own "Files changed" instead of inverting whatever landed on the base since the branch was cut. Pinned by a test.
- **The base is remembered per branch, not per repo.** A feature branch cut from `main` and one stacked on that feature branch merge into different places; a single per-repo slot would mislabel every second checkout.
- **Remote bases beat local ones.** The default is `origin/HEAD`'s target, then conventional trunk names, preferring `origin/main` over a local `main` that in a long-lived clone is months stale. The picker lifts that branch to the top of its section.
- **`Behind (N)`** reports the base's last *fetched* state. Nothing here fetches on its own.
- **`-z` record parsing** for `--numstat` / `--name-status`, so paths with spaces or non-ASCII bytes arrive unquoted and match the ones the diff overlay addresses.

The pane stays read-only: opening the pull request still happens in the terminal.

Release Notes:
- The git pane has a new **Compare** tab that shows what the current branch would merge into another — the files and commits a pull request from it would carry, diffed from the merge base so the list matches what GitHub shows.

### DIFF
--- a/Sources/termio/Git/GitChangesView.swift
+++ b/Sources/termio/Git/GitChangesView.swift
@@ -63,6 +63,7 @@ struct GitChangesView: View {
             topBar
             switch mode {
             case .changes: changesBody
+            case .compare: GitCompareView(model: model, repoRoot: repoRoot, chrome: chrome, font: settings.interfaceFont)
             case .history: GitHistoryView(model: model, repoRoot: repoRoot, chrome: chrome, font: settings.interfaceFont)
             }
             // Only Changes has a real total to report ("N files +A −D"). History's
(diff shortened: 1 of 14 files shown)


jiweiyuan:
Codex review found five edge cases where the comparison showed a wrong answer rather than a broken one — stale overlay on base switch, renames diffing as whole-file adds, no-merge-base degrading to an empty list, behind-only branches labelled "even", and a non-`origin` upstream offered as its own base. All fixed in c16bfb3 with three new tests.

PR #340, opened 2026-08-17
Judge the early client by where it gets, not when

jiweiyuan:
`attaching_mid_flood_joins_the_stream_exactly_once` has been red on the Linux runner while passing everywhere else — it was already failing on main before today's libghostty bump. Both failures read the same:

```
the early client stalled at 39895 while the late client joined at 40574 — the barrier blocked another client's delivery
```

The barrier is innocent. The early client attaches before the flood and reads it from the very first byte; the late client joins ~40k lines in and stops as soon as it has 100 lines past its boundary. Comparing the two positions at that instant asks whether the early client's *consumer* — socket, bridge process, pipe, reader thread — had caught up to a producer that never paused. On a fast machine it has; on a loaded runner it has not, and the test called that a stall.

Reproduced with no barrier involvement at all by sleeping 400µs per frame in the early reader: same assertion, same message, and `assert_consecutive` still clean — which is the tell, since the stream was intact, just short.

The fix drains the early client until it passes the boundary, with a 30s deadline. A barrier that genuinely blocked it never gets there and the deadline reports that; ordering and exactly-once stay enforced by `assert_consecutive` over everything the client received. Verified both directions — the slow-reader variant fails before this change and passes after.

Release Notes:
- None.

### DIFF
--- a/termiod/src/session.rs
+++ b/termiod/src/session.rs
@@ -1881,6 +1881,14 @@ mod tests {
         session: &mut Session,
         id: &str,
     ) -> (mpsc::UnboundedReceiver<ClientEvent>, Arc<ClientBacklog>) {
+        attach_client(session, id, true)
+    }
+
+    fn attach_client(
+        session: &mut Session,
+        id: &str,
+        snapshot: bool,
+    ) -> (mpsc::UnboundedReceiver<ClientEvent>, Arc<ClientBacklog>) {
         let (client_tx, client_rx) = mpsc::unbounded_channel();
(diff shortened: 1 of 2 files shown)


jiweiyuan:
Codex review: daemon path innocent, 30s deadline safe, but the drain loop drops finite-stall coverage — a 5s stall would now pass. Added `a_pending_snapshot_never_holds_up_another_client`, which holds the barrier open and mutation-checks that a held `Live` client fails it.

PR #341, opened 2026-08-17
Drain the early client before judging the join invariant

jiweiyuan:
`attaching_mid_flood_joins_the_stream_exactly_once` has failed twice on `main` today, both times on the Linux job while macOS passed:

```
the early client stalled at 39895 while the late client joined at 40574
— the barrier blocked another client's delivery
```

The barrier is not what it caught. The test snapshotted the early client's accumulated bytes the moment the late client was satisfied, then asserted it had already reached the late client's join counter — with nothing waiting for it in between. The early client reads the flood from its very first byte while the late one joins 40k lines in, so at any instant it is legitimately behind. The assertion measured the reader's progress, not the barrier, and a loaded runner is enough to lose the race.

Nothing in the daemon is implicated: the failing job's log carries no `resync` or `vt_stale`, so the `CLIENT_BACKLOG_CAP` path from #316 never fired, and the same commit passed on its own branch run minutes before it failed on the merge.

The fix drains the early client until it passes the boundary, bounded by a deadline. A barrier that genuinely blocked another client's delivery never gets there, so the invariant still fails — it just fails for the reason it names, with the position it actually reached. The early client also stops being killed before it has been judged, since it is the control for that half of the invariant.

Verified by CI rather than locally: the VT crate builds libghostty through `zig`, which isn't installed on my machine, and this is a load-dependent race that a single local pass would not settle anyway. The Linux job is where it reproduces.

## Release Notes

- None. Test-only; the flake never reached a build.

### DIFF
--- a/termiod/tests/join_invariant.rs
+++ b/termiod/tests/join_invariant.rs
@@ -317,11 +317,12 @@ fn attaching_mid_flood_joins_the_stream_exactly_once() {
         }
     }
 
+    // The early client keeps reading: it is still the control for the second
+    // half of the invariant, and where it has got to is judged below, once the
+    // boundary is known.
     let _ = late.kill();
-    let _ = early.kill();
     drop(frames);
     let _ = late_reader.join();
-    let _ = early_reader.join();
 
     let snapshot = snapshot.expect("no S snapshot for the late client");
     assert_eq!(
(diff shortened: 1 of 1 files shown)


jiweiyuan:
Duplicate of #340 — same fix, opened first. Closing this one.

PR #388, opened 2026-08-21
Declare the Termio requirement in the skill's compatibility field

jiweiyuan:
The `termio` skill only works inside Termio, and said so in prose at the end of its `description`. The [Agent Skills spec](https://agentskills.io/specification) defines `compatibility` for exactly this — environment requirements, max 500 characters — so the requirement is now stated where a client can read it, not only where a model can.

It is added rather than moved. `description` is the part loaded at startup for routing, and `compatibility` support varies between agent implementations; dropping the sentence would trade a signal that works everywhere for one that might not be read at all.

All three mirrors — the bundle resource, `skills/termio/`, and `web/landing/public/skill.md` — stay byte-identical, so `SessionSkillInstallerTests` still passes. `swift build` and `swift test` are green (441 tests, 0 failures).

Release Notes: None


jiweiyuan:
Dropping this — the compatibility field isn't wanted here.

PR #404, opened 2026-08-22
Make the daemon the only session backend

jiweiyuan:
Local sessions now run in `termiod` on macOS too, and the in-process PTY is gone. One path: a Unix socket for this Mac, SSH stdio for a VPS, the same framed protocol on both.

Net **−1054 lines**.

## What had to be built first

**The write token could not move.** It was assigned by attaching (`session.rs`, `AddClient`), so whichever client attached last silently muted the others and the only way to answer was to tear the attachment down and rebuild it. A phone opening a session made the Mac read-only. `claim_writer` makes the token follow the device being *used* — a client asks for it when its user types, the same rule `PTYProcess.claimHostOwnership` followed. Observers are refused: they attach without a tty and cannot resize, so handing one the token would strand the session at the last real client's size.

**`request_snapshot`** covers the one gap the daemon genuinely cannot see: frames dropped on the Mac's *own* outbound socket to a slow phone, downstream of everything the daemon's backlog protects. The bridge asks to be repainted instead of wiping and hoping.

Both verbs are additive — an older daemon decodes them to `Control::Unknown` and ignores them.

## The companion was broken with the daemon on

`companionPTY` returned `ptyProcesses[session.id]`, which is empty on the daemon path, so every phone attach was refused. The phone is now a second client of the same session rather than a tap on the Mac's copy of it.

Three things the bridge did by hand are the attach itself now, because the daemon renders its snapshot at *this client's* grid:

- the 128 KB replay cap, which existed because replaying the Mac-width ring at a phone's narrow grid reflowed it all at once and could tip libghostty's allocator into its panic screen
- skipping that replay for an alt-screen TUI, plus the mode preamble to restore those modes afterwards — a snapshot carries modes, charsets and the scrolling region
- the wipe-and-jiggle dance around a foreign winsize, now a keyframe on every resize barrier

159 lines of bridge become 100.

## What was deleted

`PTYProcess.swift` (1110 lines), `TERMIO_TERMIOD` and all eight branches on it, the `ptyProcesses` table, the flag-off placeholder surface and alerts, and `PTYForegroundTests` — it covered the in-process live-member search, whose counterpart in `proc.rs` has its own tests.

`reapStrayOrphans` survives as `StraySessionReaper`. The owner it defends against is now `termiod` rather than this app, but the leak has the same shape: a session whose owner died without teardown re-parents to launchd, keeps its dead PTY, and idles forever because agent TUIs swallow SIGHUP. A tombstone records that a session *ended*; it does not reach the process.

## Verification

- `cargo test` — 97 unit plus every integration suite, including `join_invariant` and `graceful_shutdown`
- `swift build`, `swift test` — 492 passed
- Against a real daemon: `TermiodFiles`, `TermiodWireOrder`, `TermiodWriteToken` — 10 passed
- The write-token test was checked in both directions: with the claim disabled it fails on exactly the assertion that the Mac never gets the token back

## Not covered by any of that

- **No screen verification.** Terminal correctness mostly shows up only on screen, and this touches the surface builder. Prompt reflow, alt-screen entry, and resize repaint need a real run.
- **No phone verification.** The companion rewrite has not been exercised against a device.
- **Rollback is a release rollback**, not a flag — that is what removing the fork costs, and it is why the RFC wanted a full release on the daemon backend first. That soak did not happen.

Release Notes:
- Sessions now run in the termiod daemon on macOS as well, so quitting Termio detaches instead of killing them.


jiweiyuan:
`removeProject` drops `ptyProcesses[…].terminate()` with no `termiodLinks[…]?.killAndClose()` beside it, so removing a project now leaves every agent in it running in the daemon and leaks the attach socket — while `closeSession` and `relaunchSession` two functions away both kill; `scripts/build-app.sh`'s dev-channel skip and AGENTS.md's "no zig toolchain is needed" also still describe a world where the backend is opt-in.

jiweiyuan:
Fixed in b8d7198: `removeProject` now calls `killAndClose` like the other destroy paths, with `TermiodDestroyIntegrationTests` pinning all three verbs (remove project, close session, quit-must-not-kill) against a real daemon — verified red without the fix.

jiweiyuan:
Second one, same shape: `makeStatusTap` samples `lastInputAt` off the Mac's link, but the phone now types into its own link, so `noteUserInput` never fires for phone input and the keystroke-echo guard in `noteOutputActivity` stops holding — two seconds of composing on the phone promotes an idle agent to `working` — while both doc comments still name the companion bridge as crossing that one choke point.

jiweiyuan:
Fixed in 5c5d8ce: the input clock moves from the link to the store, so Mac keystrokes, the phone bridge and `sessions send` all stamp one session-level timestamp — which is what the guard's own comment already claimed. `SessionInputClockTests` pins all three routes plus the unknown-wire-id case; verified red by severing the phone route.

jiweiyuan:
Third one, introduced by the second fix: `makeStatusTap`'s `lastInputAt` closure now reads `@MainActor TermioStore.lastInputAt` from the daemon reader thread (instrumented the read site — 1 of 1 reads off-main), so a clock that was deliberately lock-guarded on `TermiodSessionLink` is an unsynchronised dictionary raced against `noteInput` on every keystroke; and deleting `bridge.onInput = { … }` leaves all three `SessionInputClockTests` green, so nothing pins the phone route the fix added.

PR #406, opened 2026-08-22
Record ghostty's present and layout paths behind a dev-only probe

jiweiyuan:
A terminal pane flashes once while an agent is working — the whole surface, text unmoved, most visible where the text is dense. It is intermittent, comes in bursts, and does not happen in Ghostty itself. It has outlived two rounds of source reading because nothing in the app could say what actually reached the screen.

ghostty presents through a `CALayer` subclass it registers at runtime as `IOSurfaceLayer`, by two paths that carry no frame number: the renderer thread hands each surface to the main queue as a block, while the layer's own `display` pass renders inline on the main thread and assigns `contents` directly. Either can put an older frame on screen after a newer one, and which happens depends on how deep the main queue is at that moment.

## What it records

One timeline, written to `/tmp/termio-dev-present-trace.log`:

- every present, with the **raw identity** of the surface that landed, its size, which path it came by, and the gap since the last one
- every inline `display` pass — the only path that can break frame order
- every `layout` and `setFrameSize` on a terminal view

That last one is not a guess about where to look. Two other apps embedding libghostty traced their own terminal flicker to exactly that churn — manaflow-ai/cmux#2279 states plainly that it *"does NOT happen in upstream Ghostty — this is a cmux-induced bug, not a libghostty issue"*, and cmux#1625 blames `needsDisplay` set too broadly and SwiftUI state changes re-rendering terminal views. termio has the same shape: `TerminalPane` observes the whole store, so any store change re-renders every mounted surface.

## Deliberately no live frame-order check

An earlier version inferred the swap chain's cycle live, by naming its three targets in the order their surfaces first appeared. That is only the cycle when no two presents land close enough together to arrive in either order — and they routinely do. It produced three flavours of false positive and found nothing real: a permuted cycle after a resize reading as 65% out of order, a warm-up burst reading as 99%, then an unlabelled present shifting every later label by one.

The trace carries surface identity, so the order can be rebuilt exactly and offline with no premise at all. On a clean capture that analysis shows 0/42 and 0/22 presents deviating from a one-slot advance. Only a target rebuilt at a *new size* still raises a dump, because that needs no inference to recognise.

## Cost

Gated to the dev channel, and **nothing is interposed until a trace is armed from the palette** — the dev channel is the build every other piece of UI work happens in, and one carrying live interposition all the time could contaminate an unrelated investigation.

Measured, not asserted:

| state | output |
| --- | --- |
| not armed | 0 bytes |
| armed, idle | ~2 lines/s |
| armed, presenting at 120 Hz | ~360 lines/s, hard-stopped at ~20 MB |
| a dump | built once, written once; automatic ones stop after 8 |

## What it has found so far

Not the bug. Frame reordering is ruled out on clean data, and the inline render does not fire during ordinary output. Two signals remain unexplained: a run that logged more than two hundred `display` passes, and a real one-pixel resize (`1267x1416 → 1266x1416`) with a display pass at it — the geometry-jitter class of cmux#3051 and ghostty#12161.

Landing the instrument now so the next capture is against a stable, bounded version of it rather than a fourth revision.

Release Notes:
- None. Dev-channel diagnostics only, inert unless explicitly armed.

### DIFF
--- a/Sources/termio/App/App.swift
+++ b/Sources/termio/App/App.swift
@@ -138,6 +138,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
         // Dev-only: focus-independent window snapshot for automated UI debugging
         // (see DebugWindowSnapshot).
         if AppChannel.isDev { DebugWindowSnapshot.installTrigger() }
+        // Dev-only: registers the present probe's triggers. It stays inert — and
+        // interposes on nothing — until a trace is started from the palette.
+        TerminalPresentProbe.install()
         // Sweep up session processes a previous instance stranded (crash,
         // force-quit, dev rebuild's kill -9) before this run adds its own.
         StraySessionReaper.reapStrayOrphans()
--- a/Sources/termio/CommandPalette/CommandPalette.swift
+++ b/Sources/termio/CommandPalette/CommandPalette.swift
@@ -405,6 +405,25 @@ struct CommandPaletteView: View {
                                  icon: nil, symbol: "square.3.layers.3d", shortcut: nil) { _ in
                 NotificationCenter.default.post(name: .termioDebugDumpLayers, object: nil)
             })
+            // Writes the last frames ghostty put on screen — slot, path (async vs
+            // the layer's inline `display` render), and the gap between them — to
+            // stdout and /tmp/termio-dev-present.log. Run it right after a flash:
+            // a slot that does not advance by one is a frame that landed late.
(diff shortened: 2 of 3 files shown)


jiweiyuan:
Draft until a foreground capture either confirms the 1px width oscillation (`1266↔1267`) — in which case this trims to the size-change and invalidation halves and lands with the fix — or comes back clean, in which case it closes in favour of a docs/bug handoff.

PR #410, opened 2026-08-22
Tell a lost connection apart from a session that ended

jiweiyuan:
`handleStreamEnd` delivered `exit 1` when the transport died. The child never produced that 1 — it is almost certainly still running, which is the entire point of it living in a daemon — and the invented status then went to `sessionExit`, which parks the pane over an error with no error output behind it.

So pulling the network, restarting the daemon, or dropping an SSH pipe made a VPS agent that is working fine read as one that crashed.

## The split

They are different events and only one of them carries a status. `TermiodSessionLink` gains `onConnectionLost` beside `onExit`; the store drops the dead attachment and says so on screen — a pane that merely stops updating is indistinguishable from an agent that went quiet — but runs **no exit policy**, keeps the row, and leaves the session alone. Selecting it again attaches by name to the same process.

Deliberately not attempted here: automatic reconnection. That belongs with `TermiodConnection` (RFC Stage 4b), which owns transport health. This change only stops the lying.

## Why now

This is the one failure mode with no fallback left after #404 deleted the in-process backend — the daemon is the only thing holding the PTY. It is also the prerequisite for supervision (Stage 5): you cannot sensibly restart a daemon whose death you cannot distinguish from a process exiting.

## The tests found the distinction they needed

`TermiodFaultIntegrationTests` is the first coverage of this mode at all — every other daemon suite is a happy path that never kills anything.

Writing it surfaced a correctness point worth recording: a **SIGTERM is caught**, and the daemon drains — killing each session and reporting the real `session_exited` (status 137) before closing the socket. That is right, and it is not this bug. The test uses **SIGKILL**, so the daemon never gets to speak and the client sees only EOF, which is what a cut network actually looks like.

Both directions are pinned, so the split cannot degenerate into muting exits:
- killing the daemon outright reports a lost connection and **no** exit
- a genuine `exit 7` still arrives as 7, and the EOF behind it is not re-reported as a disconnection
- a session survives its connection: reattaching by name resolves the same pid rather than spawning a replacement

Verified in reverse — restoring `deliverExitLocked(status: 1)` fails exactly one test, on `a broken transport was reported as the session exiting with [1]`.

## Verification

- `swift build`, `swift test` — 515 passed
- Daemon suites against a real `termiod` — 129 passed
- `scripts/check-strings.sh` passes; the two new strings carry zh-Hans, and the interpolated one is keyed `%@` to match how `String.LocalizationValue` compiles it

Release Notes:
- A dropped connection to a session now says the connection was lost instead of reporting the session as having exited. The session keeps running and can be reattached.

### DIFF
--- a/Sources/termio/Resources/Localizable.xcstrings
+++ b/Sources/termio/Resources/Localizable.xcstrings
@@ -5994,6 +5994,36 @@
           }
         }
       }
+    },
+    "this Mac": {
+      "localizations": {
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "本机"
+          }
+        }
+      }
+    },
+    "Select it again to reattach.": {
+      "localizations": {
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "重新选中它即可接回。"
+          }
+        }
+      }
+    },
+    "Lost the connection to %@. The session is still running there.": {
+      "localizations": {
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "与 %@ 的连接已断开。会话仍在那台机器上运行。"
+          }
+        }
+      }
     }
   },
   "version": "1.0"
--- a/Sources/termio/Resources/Localizable.xcstrings
+++ b/Sources/termio/Resources/Localizable.xcstrings
@@ -6005,22 +6005,22 @@
         }
       }
     },
-    "Select it again to reattach.": {
-      "localizations": {
-        "zh-Hans": {
-          "stringUnit": {
-            "state": "translated",
-            "value": "重新选中它即可接回。"
-          }
-        }
-      }
-    },
     "Lost the connection to %@. The session is still running there.": {
       "localizations": {
         "zh-Hans": {
           "stringUnit": {
             "state": "translated",
             "value": "与 %@ 的连接已断开。会话仍在那台机器上运行。"
+          }
+        }
+      }
+    },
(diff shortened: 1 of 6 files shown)


jiweiyuan:
The notice does not survive a render: `TerminalPane`'s `ForEach` calls `store.surface(for:)` for every mounted pane on every render (`activated` never drops a session), so clearing `surfaces[id]` rebuilds unbidden rather than on reselect — and with the transport still down `start()`'s catch delivers `deliverExitLocked(status: 1)` into `applyTermiodExit`, parking the pane over the invented exit 1 this PR exists to remove; reproduced by killing the daemon and then calling `surface(for:)` once more.

jiweiyuan:
The widened catch now covers the daemon answering as well as failing to reach it: with a healthy socket and a cwd that no longer exists the daemon returns `CreateFailed`, `start()` throws `handshakeRejected`, and the pane says "Lost the connection to this Mac. The session is still running there." — connection fine, nothing running, and closing and reopening will fail identically; the two cases are already distinguishable at the throw site.

jiweiyuan:
Reviewed and merging. Fixed one thing in b27515f rather than blocking on it: the refusal arm was throwing away the daemon's reason — every non-`.attached` reply collapsed into `handshakeRejected("attach was not acknowledged")`, so a missing cwd printed that instead of the device's "spawning session program ... No such file or directory", and the existing non-empty assertion could not catch it. Noting without fixing: the three notices are still unseen on screen, reattach is still untested end to end, and `deliverStartRefusedLocked` landed inside `deliverConnectionLostLocked`'s doc comment.

PR #414, opened 2026-08-23
Pick a workspace's colour from the theme's own palette

jiweiyuan:
Follows #412, which gave a workspace a colour and put it in the switcher, and #413, which built the pane that lists every workspace at once.

## Twelve hues, all the theme's

The palette was five bright ANSI slots. It is now six hues at two intensities — bright (9–14) over normal (1–6) — so a picker six to a row reads as **one hue per column, one intensity per row**. Still nothing invented: a colour that came out of a palette of ours would be a stranger the moment the user changes theme, which is the same reason the stored value is an index rather than a colour.

Black and white are left out at both intensities. Those are the slots a theme spends on its own background and foreground, so as a mark they read as no colour at all.

## The picker

26pt swatches spread to the width of the name field with `.equalSpacing`, rather than bunched at its leading edge — a short last row keeps its columns under the full one.

Selection is a **ring around a gap around a smaller dot, all in the swatch's own colour**. That is what this system's own pickers do, and what Dia builds out of a `mainLayer` and a separate `selectionLayer`. The previous treatment was an outline in the label colour, which cannot work against a palette taken from the user's terminal theme: whatever fixed lightness the outline has, half the palette will swallow it. A shape made of the swatch's own colour cannot lose contrast with itself.

## And in Settings ▸ Workspaces

The mark is drawn on every row and can be changed there, because that pane is the other place the whole set is visible at once.

It is a submenu inside the row's existing overflow menu, not a second control beside the name: that row's shape is one control at the trailing edge and type everywhere else, and a colour well would put a second chrome inside a grouped row that already draws its own background. The current colour carries a tick, so it is never signalled by colour alone.

`setWorkspaceColor` is its own store verb — the pane changes a colour without touching the name, and routing it through `renameWorkspace` would make an unchanged name look like an edit to every observer.

## Also carried

Two pieces of the same afternoon's work on the switcher that live in the same files: the Settings window takes the terminal theme's background under a transparent titlebar the way the main window does, and the menu's own verbs align with the rows above them now that those rows carry a swatch.

## Verified

`swift build`, `swift test` (522 tests, 0 failures), `check-strings.sh` clean. Driven by hand on a dev build: creating a workspace, changing a colour from the settings pane, and watching the switcher's dots follow.

Release Notes:
- A workspace's colour can be picked from a wider palette, and changed later in Settings ▸ Workspaces.

### DIFF
--- a/Sources/termio/App/App.swift
+++ b/Sources/termio/App/App.swift
@@ -812,6 +812,18 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
         NSApp.appearance = appearance
         window?.appearance = appearance
         settingsWindow?.appearance = appearance
+        applySettingsWindowChrome()
+    }
+
+    /// Paints the Settings window in the terminal theme, the same recipe the main window
+    /// uses: an opaque themed background under a transparent titlebar, so the toolbar band
+    /// sits flush with the panes instead of on AppKit's stock grey. Re-run whenever the
+    /// theme changes (`applyChromeAppearance` is on the settings observer), because the
+    /// window paints this itself and won't follow the SwiftUI tree.
+    private func applySettingsWindowChrome() {
+        guard let settingsWindow else { return }
+        settingsWindow.backgroundColor = resolvedTerminalBackground()
+        settingsWindow.titlebarAppearsTransparent = true
     }
 
     /// Installs a real `NSToolbar` carrying CodeEdit's chrome: a leading navigator toggle, the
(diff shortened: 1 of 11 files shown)


jiweiyuan:
Dropping the workspace colour. It answers a question the row's own name already answers: termio is single-window, the scope is named in the toolbar and the sidebar lists only that scope, so there is no ambient confusion for a colour to prevent — which is the whole reason a browser tints a profile. The one high-cost confusion here is which *machine* a session is on, and that is a fact the device mark already carries.

PR #446, opened 2026-08-25
Scope settings to the machine they mean

jiweiyuan:
Settings silently meant *this Mac*. `AgentAvailability` probed this Mac's `PATH` to decide whether an agent was installed, `agents.commandOverrides` held one value per agent with no machine dimension, and the hook and skill installers wrote into this Mac's `$HOME`. Switching to a VPS workspace changed nothing on the page.

Implements `docs/design/20260824-settings-that-know-which-machine.md`, D2 and D4 and D6 through D8. Device tint (D3) and Mobile as a Serving section (D9) are deliberately not in this pass.

## D2, the blocker, and what was decided

`SettingsTab.swift` was `case devices = "ssh"` — the shipped Devices tab was the renamed SSH tab, a `~/.ssh/config` projection, which `20260814-remote-to-device.decisions.md` §1 classifies as *routes*. Machine identity therefore had no home and the name it wanted was taken. #440 landed while this was in flight and made that tab a substantially richer route editor (Keychain passwords, `ssh-copy-id`, credential grouping) without adding any identity surface, which strengthens rather than changes the answer.

**Decision: one tab, called Machines, one row per machine, and the route/identity distinction becomes two sections inside a machine's pane** — *Reached by* (destination, key, Test Connection, Edit in Config) above *Runs* (agent CLIs and their per-machine paths). Two tabs that both list machines would force the user to learn that distinction before they could find anything; one pane that shows both does not.

What stays at tab level is what is genuinely not per-machine: `~/.ssh/config` itself, and the public keys in `~/.ssh`. The raw value stays `ssh`, so `lastOpenKey` survives its third renaming.

`DevicesSettingsTab.swift` becomes `SSHHostControls.swift` — the shared route pieces plus the Add Host sheet the File menu also opens.

## What moved (D8)

| Was | Now |
| --- | --- |
| General ▸ Agent skill, General ▸ Status | the machine's pane, as installers |
| the two switches behind them | Agents ▸ Integration — wanting the feature is a preference |
| General ▸ Command line | This Mac's pane, beside deploy-`termiod` |
| Agents ▸ per-agent command path | the machine's pane |

General is now language, notifications and GitHub, and the tab stops lying about its scope.

## Storage (D7)

Authored values go in a `devices` section of `settings.json`, through the same `SettingsStore` that already writes only the keys the user set. Discovered state goes in `~/.termio/devices/<host_id>.json` — a cache, named one from the day it is written, safe to delete at any moment and never read as authority. Keying is `deviceID ?? alias ?? "local"`, the bootstrap/stable split `KnownDevice` and `Checkout.deviceIdentity` already use, so one box reached three ways keeps one blob.

Existing `agents.commandOverrides` migrate onto this Mac — the only machine they could ever have described — and the legacy key is cleared as it is read.

`resolveLaunch` now resolves the command against the machine the session runs on, so a per-machine path actually takes effect instead of handing a Homebrew path to a Linux VPS.

## Three readiness states (D4)

| State | Row shows |
| --- | --- |
| available | the command alone |
| missing | `Not installed · <command>` + `(!)` |
| **unknown** | `Can't check on vps · <command>`, no badge |

Reachability is asked once per machine rather than once per agent, so an unreachable box makes every agent unknown in one step instead of paying an ssh timeout eight times. The Agents tab's line is deliberately **passive** — it reads the device file and never fires ssh — because a tab that probed every configured host on open would hang on a sleeping VPS to decorate a line nobody asked for.

## One outcome, not a ladder (D6)

A machine's pane shows *Ready* or *Set up this device*. One click runs deploy `termiod` → probe agent CLIs → install hooks and skill, and surfaces only the **first** blocking failure: a box that does not answer also has no agent CLIs and no hooks, and naming all three invites fixing a consequence instead of a cause. The rungs remain individually runnable underneath as disclosure. This is the first caller to pass a target to `AgentConfigStore`.

## Verification

- `swift build` clean; `swift test` 632 tests, 0 failures (11 new in `DeviceSettingsTests`).
- `check-strings.sh` reports zero missing keys. The second commit adds the seventeen `%@` keys the checker structurally cannot see — every scoped line names its machine, so all of it is interpolated, and it would otherwise have shipped English to Simplified Chinese users while passing the gate.
- **Not verified on screen.** Screen Recording is denied to this environment, so `screencapture` fails and I could not read the layout back. Spacing, alignment and the drill-down chrome need a look on a real display before merge.

Release Notes: Settings now say which machine they mean. The Devices tab becomes Machines: one row per machine, opening to how it is reached and what it runs. Each machine gets its own agent command paths, its own hooks and skill install, and one "Set up this device" button that does the whole chain. An agent on a machine Termio can't reach now reads "Can't check" instead of "Not installed".

### DIFF
--- a/Sources/termio/Resources/Localizable.xcstrings
+++ b/Sources/termio/Resources/Localizable.xcstrings
@@ -61,6 +61,16 @@
         }
       }
     },
+    "%@ is ready.": {
+      "localizations": {
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "%@ 已就绪。"
+          }
+        }
+      }
+    },
     "%@ isn’t a UTF-8 text file.": {
       "localizations": {
         "zh-Hans": {
@@ -77,6 +87,16 @@
           "stringUnit": {
             "state": "translated",
             "value": "%@ 暂不支持这台设备。"
+          }
+        }
+      }
+    },
+    "%@ isn’t installed on %@": {
+      "localizations": {
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "%1$@ 未安装在 %2$@ 上"
           }
         }
       }
@@ -531,6 +551,26 @@
         }
       }
     },
+    "Agents on %@ can run, and report their status back here.": {
+      "localizations": {
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "%@ 上的 Agent 可以运行，并把状态报告回这里。"
+          }
+        }
+      }
+    },
+    "Agents on %@ can run.": {
+      "localizations": {
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "%@ 上的 Agent 可以运行。"
+          }
+        }
+      }
+    },
     "Agents won’t show as working.": {
       "localizations": {
         "zh-Hans": {
@@ -698,6 +738,16 @@
           "stringUnit": {
             "state": "translated",
             "value": "应用"
+          }
+        }
+      }
+    },
+    "Asking %@ what it has.": {
+      "localizations": {
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "正在询问 %@ 上有什么。"
           }
         }
       }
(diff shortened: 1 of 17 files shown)


jiweiyuan:
Screenshot `25-session-control-settings.png` still shows the old General pane and needs regenerating against Agents ▸ Integration; its alt text is left describing the current image rather than the new UI.

PR #472, opened 2026-08-25
Install agent integration through the daemon

jiweiyuan:
Stage 4 of [`docs/design/20260825-agent-integration-moves-to-termiod.md`](docs/design/20260825-agent-integration-moves-to-termiod.md). Swift calls the daemon; `AgentConfigStore` and `SSHAgentConfigStore` are deleted; `DevicePaneModel.setUp` is one call.

## The gate: counted, not felt

An `ssh` wrapper on `PATH`, plus an instrumented throwaway checkout of `main` for the before, both against `ukvps`:

| | before | after |
| --- | --- | --- |
| install, 12 agents | **98** ssh invocations | **1** |
| agent readiness probe | **45** | **1** |

The probe is in here because deleting `SSHAgentConfigStore` would otherwise have left the readiness rung doing one `ssh` per agent. `probe_agents` is a read-only verb that answers for the whole roster.

All twelve agents came back installed in the one round trip:

```
INSTALL ok: ["Claude Code", "Codex", "OpenCode", "Pi", "Amp", "Droid",
             "Cursor", "Copilot", "Kimi", "Qwen Code", "Cline", "Grok"]
INSTALL failed: []
```

## The three that aren't mechanical

**1. Installing on this Mac is now IPC.** `Transport.open` already auto-starts `termiod serve` when nothing answers — the same spawn every local session depends on — so "the daemon was not running" is a state this recovers from rather than one it reports. What it cannot recover from (no daemon binary at all) is logged and surfaced, never swallowed. The background callers (launch, preference change, refocus) have nowhere to show a failure, so they log it and leave the previous state alone; the caller that *does* report is Settings, where the user asked. A Mac that could always install its own hooks does not start failing quietly.

**2. Protocol skew refuses by capability, and the foundation rung is the fix.** The install negotiates `agents` at the handshake, so an old daemon produces a sentence rather than a no-op. Verified live against ukvps's genuinely old daemon:

```
INSTALL failed: ["termiod on this machine is too old to install agent integration."]
```

The device pane's foundation rung now redeploys a daemon that lacks the capability, so the rung the user was already pressing is the answer.

**What it deliberately does not do is restart a running daemon.** `mv` over a running binary leaves the process holding the old inode — that is exactly why the deploy is safe — so the new binary only takes over when the old daemon exits. Restarting it would kill every session it owns, and detach-≠-kill is the promise the daemon exists for. ukvps had 15 live sessions during this work, so this is not hypothetical. An auto-`pkill` would also have been wrong twice over: it would have hit the unrelated `--wss` daemon from #344 running on the same box. So that state is named with its cost — "close the sessions on `<host>` and set it up again" — instead of taken.

**3. D4's precondition stays, in its cheap local form.** What went with the SSH arm is `expect_sha256`: shipping a digest across a network so a *remote* check-and-swap could be atomic. That existed because the merge and the file were on opposite ends of a link, and it goes with the link. What it was protecting does not: the user's editor is on this machine and `~/.claude/settings.json` is a file people keep open. The window is now microseconds inside one process instead of two round trips, which makes the check nearly free rather than unnecessary — one extra `read` per file against silently discarding somebody's edit is not a trade worth taking.

## One protocol shape worth reviewing

`install_agents` states each half as `install` / `remove` / `leave` rather than carrying one `enabled` flag. The two Integration switches are independent, and a single flag would have meant either two round trips when they disagree, or "Reinstall hooks" quietly removing the skill when live status happens to be off. This is what keeps one message being one message.

## The split of HookListener.swift

Only the installer half left. The socket that receives reports stays, and so do the two things that were living in those files without belonging to them:

- **`HookDialect` is schema**, read by `AgentDefinition.swift`'s manifest parsing — moved next to the rest of the hook schema.
- **`XDGBaseDirectories` has one consumer left**: `AgentSessionStore`, finding an agent's own session records to resume a conversation. That is a *read* on this Mac and stays here; the install-time spelling of the same rule already lives on the box in `agent::machine::expand`.

#463's `daemonMarker` and `isTermioCommand` are ownership fingerprints used only by the installers, so they went with them — termiod has carried the same fingerprint since Stage 2, and its tests cover it.

The fixture comparing the two manifest parsers now says in its doc comment that it is a permanent contract, not migration scaffolding: Swift did not stop parsing manifests, it stopped writing from them.

## Settings surfaces — what I verified, and what I did not

The full diff across `AgentSettingsTab`, `DevicePane` and `SettingsControls` is **closure bodies only**: no view, label, title, section, or ordering changed, and `SettingsControls` has no diff at all. `InstallButtonRow` is untouched, so the spinner still runs for the length of `await action()` — the action is still async, and the `Task.detached` wrappers around the old blocking batches are gone from every Settings call site. `DeviceReadiness`'s install rung no longer detaches either, because there is no blocking work left to detach from.

**I did not get a screenshot.** The dev app built and launched from this worktree but comes up with zero AX windows and app-level AppleScript hangs — a worktree bundle is a new app to macOS's privacy system. I stopped rather than keep burning time on it, and I would rather say so than claim a visual check I did not make. If you want the visual confirmation, a screenshot of Settings ▸ Agents and a device pane from your own build would close it.

## Verification

- `swift test` — 651 pass (the ~30 installer tests moved to termiod in stages 2–3, where they are covered by unit tests and the byte-identical gate)
- `cargo test` — 157 pass
- the ssh counts above, on `ukvps`
- ukvps left correct: the current daemon binary deployed, its hooks naming `~/.local/bin/termiod`, and all 15 sessions still running

Release Notes:
Setting up agent integration on a device is now a single request to that machine instead of one connection per file — a twelve-agent install went from 98 SSH round trips to one. A device running an older `termiod` says so and is updated, rather than appearing to succeed and installing nothing.


jiweiyuan:
Both review points are in.

**The missing JSON test** — `a_second_json_install_replaces_the_first`, alongside the TOML and plugin ones. You were right that it was the gap that mattered: plugin and TOML are recognised by markers the dialect owns, JSON only by its command, which is why JSON is the one that actually doubled.

Verified by reintroducing the bug — dropping `DAEMON_MARKER` from the ownership check — and confirming it fails:

```
assertion `left == right` failed: a reinstall must replace our entry, not append
a second one — the user's hook plus exactly one of ours:
  left: 3
 right: 2
```

One thing that came out of doing it that way: my first version counted *our* groups via `is_ours`, so with the ownership check broken it reported `left: 0` — a passing-looking count of zero rather than the duplicate that was really there. The primary assertion is now the total group count, which does not consult `is_ours` at all, so a broken ownership check reports the duplicate instead of hiding behind itself.

**The comment** — `syncAgentIntegration` now says this is this Mac only and must stay that way, with the reason: local IPC is affordable on refocus, and the same loop over known machines would fire an ssh at every configured VPS on every app focus, asleep ones included, each paying a connect timeout.

776 Swift tests, 161 Rust. Ready for you to merge.

PR #487, opened 2026-08-26
fix(ios): prevent companion reconnect flicker

hello532:
## Summary
- serialize WebSocket dials so foreground/network/reconnect callbacks cannot overlap sockets
- stop sending RIS on reconnect; the attach snapshot is the authoritative repaint boundary

## Root cause
iOS could open overlapping companion sockets during reconnect triggers, while each reconnect emitted ESC c before the server snapshot arrived. That cleared the visible terminal and caused repeated attach/resize streams to appear as flicker.

## Validation
- `git diff --check` passes
- changed Swift sources compile during `swift test` build
- full `swift test` could not complete in this environment; the build ended with `error: fatalError` after dependency compilation

### DIFF
--- a/ios/Sources/CompanionTransport.swift
+++ b/ios/Sources/CompanionTransport.swift
@@ -137,13 +137,10 @@ final class CompanionTransport {
     /// screen reset, auth, the attach it gates, and the grid claim the Mac's
     /// repaint is driven by.
     private func sendPreamble() {
-        // On a REconnect the terminal still shows the dead link's screen; a
-        // full reset (RIS) ahead of the server's ring-buffer replay keeps the
-        // two byte streams from interleaving into garbage. Emitted here on the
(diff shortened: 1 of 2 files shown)


jiweiyuan:
Thanks for digging into this — you're right about the RIS, and the reasoning holds even more strongly than the PR states: termiod's snapshot carries its own prologue (`SNAPSHOT_PROLOGUE` in `termiod/vt/src/lib.rs`), which deliberately stops short of RIS because the client's palette, title and scrollback aren't the host's to reset. I'd rather not take the `dialing` latch, though: it's written from the URLSession delegate queue and read on main, `died()` clears it for a superseded task, and it turns `reconnectNow()` into a no-op while a dial is hung. Closing this and landing the RIS removal separately — much appreciated.

PR #496, opened 2026-08-26
Fix remote deploy when the app is what starts it

jiweiyuan:
Setting up a machine from the app failed with an alert that named nothing:

```
Couldn't set up ukvps
Couldn't deploy termiod to ukvps.
• or build on the host and deploy with: termiod remote deploy <host> --bin <path>
```

Three causes, stacked.

**Cargo never read `termiod/.cargo/config.toml`.** Cargo discovers it by walking up from the working directory, and `--manifest-path` does not move that search. Started from anywhere but the crate — an app's working directory is `/` — the musl cross-link never saw the config that points it at the bundled rust-lld and fell back to Apple's `cc`: `ld: unknown options: --as-needed -Bstatic -Bdynamic --eh-frame-hdr`. This failed for every caller with a complete toolchain, and is why the deploy could not work from the app at all.

**`cargo` and `zig` were assumed to be on `PATH`.** An app launched from Finder inherits launchd's `PATH` — four directories holding neither cargo nor anything from Homebrew. The lookup now goes through the login shell's `PATH`, the same probe agent discovery already uses, plus Homebrew's zig prefix: brew leaves `zig` off `PATH` entirely whenever a second `zig@N` formula holds the link, the case `scripts/build-app.sh` already checks for. Each missing tool is named before cargo runs — a missing zig used to surface minutes in as a build-script panic, under advice naming the wrong problem.

**The alert kept only the last line of the output.** `remote deploy` reports through anyhow, which prints one multi-line `Error:` block — headline, then the fixes, one per line. Keeping the last line alone is what put a bare bullet under the title. The whole block is kept now, cargo's own output above it dropped.

## Verification

Run under the app's exact environment (`env -i` with launchd's `PATH`), against a real VPS:

```
[deploy] cross-compiling for aarch64-unknown-linux-musl…
[deploy] installing …/aarch64-unknown-linux-musl/release/termiod → ukvps:~/.local/bin/termiod
[deploy] installed: termiod 0.1.0
exit=0
```

Before: `Error: running cargo build (is cargo on PATH?)` immediately, or the `cc` link failure with a full `PATH`.

`cargo build --release` and `swift build` both clean on this branch.

Release Notes:
- Fixed setting up a remote machine from the app, which failed to deploy `termiod` even with a complete Rust toolchain installed.
- Deploy failures now explain what went wrong instead of showing a single stray line.

### DIFF
--- a/Sources/termio/Terminal/Termiod/TermioStore+Termiod.swift
+++ b/Sources/termio/Terminal/Termiod/TermioStore+Termiod.swift
@@ -1055,11 +1055,8 @@ extension TermioStore {
         }
         let deploy = runProcess(localBinary, ["remote", "deploy", host])
         guard let deploy, deploy.exitCode == 0 else {
-            let detail = deploy.map { output in
-                (output.standardError + output.standardOutput)
-                    .split(separator: "\n").map { $0.trimmingCharacters(in: .whitespaces) }
-                    .last(where: { !$0.isEmpty }) ?? "deploy failed"
-            } ?? "couldn't run termiod remote deploy"
+            let detail = deploy.map { deployFailureDetail(from: $0) }
+                ?? "couldn't run termiod remote deploy"
             return .failure(RemoteSetupError(
                 message: "Couldn't deploy termiod to \(host).\n\(detail)"))
         }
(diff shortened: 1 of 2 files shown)


jiweiyuan:
Superseded by #499: deploy is now the lifecycle loop, which takes the bundled slice before it ever cross-compiles.

PR #502, opened 2026-08-27
Let a termiod update through a tab that is only watching a prompt

jiweiyuan:
Follow-up to #499. `termiod stop` (and so *Set Up* in Settings ▸ Machines) declined while any client was attached — and the attached client is usually the app whose user just asked for the update, so its own idle tabs vetoed it with "close those and set it up again".

Busy is now **work in progress**: a command running in a session's foreground (`foreground_job`, the signal the daemon already computes for "closing this loses work") or an agent `working` / `needs_you`. A client on a shell at its prompt no longer holds the daemon up. When there is work, the pane names it (by title or command, not the app's UUID) and offers **Update Anyway**, which runs the loop with `--force`.

RFC §11 records the change to the idle definition.

## Verified
- `cargo test` (180 unit + integration suites) and `swift build` / affected Swift tests green.
- Against the VPS with the release app attached to an idle login shell: see the PR comment for the run.

## Release Notes
- Updating termiod on a machine is no longer blocked by your own idle terminals there; when an agent or command is still running, the pane names it and offers Update Anyway.

### DIFF
--- a/Sources/termio/Resources/Localizable.xcstrings
+++ b/Sources/termio/Resources/Localizable.xcstrings
@@ -6879,6 +6879,16 @@
         }
       }
     },
+    "Update Anyway": {
+      "localizations": {
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "仍然更新"
+          }
+        }
+      }
+    },
     "Update ready": {
       "localizations": {
         "zh-Hans": {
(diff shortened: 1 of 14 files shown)


jiweiyuan:
Verified on the VPS with the release app attached to an idle login shell: staged → stopped → verified as current in 1.7 s; the app's pane saw the shell exit.

PR #508, opened 2026-08-27
Publish a device from the pane that pairs it

jiweiyuan:
Pairing an iPhone with a Linux box required the user to already own a domain, a reverse proxy and root on that machine. The pane's only offer was a text field asking them to type the resulting address — and that field could not work: `--url` changes the invite `termiod pair` prints, not the origins the daemon accepts, so it produced a QR that scanned and then took a 403.

The address must be byte-identical in three places (the invite, the daemon's `--wss-origin`, and the origin the phone stores and sends), which is not something to ask a person to retype. So the pane publishes the box instead of asking about it: the Mac already has SSH there, runs the tunnel on the far machine, and reads the address back.

The choice is the same list the Mac's own Mobile Access offers — Tunelo, Cloudflare, ngrok, Custom — and it exists because Termio's relay has finite capacity: a tunnel the user already owns is a first-class answer, not a consolation. Custom keeps the Mac's command-plus-regex contract but is stored per machine.

Notes:

- **Nothing needs root.** `~/.local/bin`, `~/.local/state`, `systemctl --user` with linger. Requiring root is what turns "connect my phone" into sysadmin work.
- **Stability is modelled, not assumed.** The phone persists the address and origin and the daemon pins it, so a rotating name strands every paired phone. Providers that don't hold their address say so before the click.
- **Arming asks first and names the sessions** it would end, by command line — a count can't inform that choice.
- **Every remote `pkill` is anchored** to the binary path; unanchored, `pkill -f` matches the ssh wrapper's own command line and kills the shell running it.

Verified against a real VPS: Publish installs and starts the tunnel, reads back a stable subdomain, arms the daemon, and the pane draws a QR. Dialling that URL from the Mac with the headers the phone sends returns `101 Switching Protocols` with the subprotocol echoed, answering as the `host_id` the box reports.

Not covered: an actual phone scanning the QR, and the relay's `--max-session` is still the 86400s default, so a tunnel drops after 24h.

Release Notes: Settings ▸ Mobile can now publish a remote device to the internet and pair your iPhone with it — pick a tunnel (Tunelo, Cloudflare, ngrok, or your own), press Publish, and scan the code. Nothing on the device needs root, and the tunnel is set up to survive a reboot.

### DIFF
--- a/Sources/termio/Resources/Localizable.xcstrings
+++ b/Sources/termio/Resources/Localizable.xcstrings
@@ -41,12 +41,32 @@
         }
       }
     },
+    "%@ answered as a different machine than %@. Its address is being served by something else.": {
+      "localizations": {
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "%1$@ 回应的是另一台机器，不是 %2$@。这个地址正被别的东西占用。"
+          }
+        }
+      }
+    },
     "%@ answered, but not with a pairing invite. Its termiod is probably older than this app — deploy it again.": {
       "localizations": {
         "zh-Hans": {
           "stringUnit": {
             "state": "translated",
             "value": "%@ 有响应，但返回的不是配对邀请。它的 termiod 可能比本应用旧——重新部署一次。"
+          }
+        }
+      }
+    },
+    "%@ answered, but not with the handshake termiod sends.": {
+      "localizations": {
+        "zh-Hans": {
+          "stringUnit": {
+            "state": "translated",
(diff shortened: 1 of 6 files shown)


jiweiyuan:
Blocking: Custom relays can publish a path-mounted address (for example, `https://host/termio`), but `publish()` passes that full URL as `--wss-origin`. The iPhone correctly derives and sends an Origin without a path (`https://host`), so termiod rejects the connection with 403.

Derive the origin from the published URL’s scheme, host, and port before calling `arm`, while retaining the full URL in the invite. Please add coverage for this path-based case.

PR #518, opened 2026-08-29
Make the remote file tree live over the fs: resource

jiweiyuan:
The remote file tree re-listed every directory it was showing on every app focus. On a box where an agent is writing files, that is a full round of `fs.list` each time you come back to the window — to learn that nothing you can see has changed.

`termiod` has served the `fs:` resource since it grew a watcher (`a6a1204`, shipped in v0.37.0). Nothing on the Swift side ever subscribed; `TermiodFiles.swift` said so in a comment.

## What was missing

**The pool dropped every frame that answered no request.** A request's answer belongs to its inbox and dies with it; a resource's batches belong to the *channel* and outlive every request on it. `PooledChannel` grew a second delivery path (`addObserver` / `ChannelSignal`) alongside the inboxes.

**The tree had no way to apply a delta.** It does now. A batch names the directories whose contents moved; only the ones the tree has actually realized are re-listed, in one request. A batch that touches nothing on screen costs nothing at all — the rule VS Code's explorer uses, where a file event refreshes only when it hit a visible item.

The whole-tree reload is still here and still one request for every open directory, but it is the exception now rather than the beat: a first load, a `full_rescan` batch, a dropped subscription, and the refresh button.

## Two details the wire forces

**The host canonicalises the resource id.** `daemon.rs`: *"Clients name a workspace root; the host canonicalises it, so two spellings of one repo share a single watch."* A root reached through a symlink comes back under a prefix no row in the tree carries. The watch adopts the id the host answers with, and the tree translates batch paths back into its own spelling — every path it holds came from an `fs.list` reply, which echoes the path as asked.

**A first subscribe always reports `gap`**, having nothing to replay from. That is not news to a caller that is loading anyway, so it is reported only on a re-subscribe; otherwise every pane opened with two full listings instead of one.

Both were caught by the integration tests, not by reading.

## Measured

| | |
| --- | --- |
| batch after one write, local daemon | 314 ms |
| batch over SSH to a VPS | 507 ms |

Both are the daemon's own 0.3 s quiet window plus the link. Against a real checkout the first batch after subscribe named 200+ directories, which is the filter earning its keep: with only the root realized that is one round trip, not two hundred.

## Testing

- 860 tests pass, with and without a real daemon.
- New unit tests on the relist filter: visible-only, dedup, collapsed folders, changes under an unopened subtree.
- New integration tests against a live daemon: batch delivery names the changed *directory*, and a listing taken under a watch carries that watch's cursor.
- Verified over SSH against a real box: the subscription resolves, and a remote `touch` lands in 507 ms.

Not verified on screen — the claim that a file created on the box appears in the tree untouched has no screenshot behind it.

Release Notes: The file tree for a remote project now updates itself when files change on that machine, instead of reloading when you switch back to the window.

### DIFF
--- a/Shared/Sources/TermioShared/TermiodProtocol.swift
+++ b/Shared/Sources/TermioShared/TermiodProtocol.swift
@@ -30,7 +30,7 @@ public enum Termiod {
     /// | `scrollback` | no      | `H` carries packed cells to inject *above* the viewport; a byte-stream surface has nowhere to put them |
     /// | `grid_diff`  | no      | `G` would make the host resolve every cell's colour, which overrides the viewer's theme — the §A/§H regression this client exists not to repeat |
     /// | `send_wait`  | no      | `send`/`wait` are control-channel verbs; the app injects through its own attach channel |
-    /// | `resources`  | no      | `subscribe_resource` — live tree and git updates. The channel that outlives a request now exists (`TermiodControlPool.swift`); what is still missing is a tree that applies deltas rather than re-listing |
+    /// | `resources`  | no      | `subscribe_resource` — the live file tree (`Termiod.ResourceWatch`), on the same pooled control channel the listings ride, never an attachment |
     /// | `fs_watch`   | no      | ditto |
     /// | `files`      | no      | `fs.list`/`fs.read` — the Files pane's consumer, on the device's pooled control channel (`TermiodFiles.swift`), never an attachment |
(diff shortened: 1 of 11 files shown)


jiweiyuan:
Both review findings fixed in 0cf2104: the watch now reports `established` so a tree listed before any watch existed re-reads once, and the app-focus reconcile is back for daemons that never grant `resources`.

PR #522, opened 2026-08-29
Replace the termiod binary without dropping sessions

jiweiyuan:
Upgrading a box used to cost the work running on it. `termiod deploy` staged the
new binary, asked the daemon to stop, and every PTY master closed with the
process — SIGHUP, and the agents mid-task went with it. A box with work in
progress was left permanently `staged` instead, waiting for someone to decide
whose session to end.

`execve` is the way out. It replaces the process *image* while keeping the
*process*: same pid, same children, same open descriptors, same session and
process group. The PTY masters stay open, the shells never see a hangup, and
the only thing that changes is which code is on the other end of the descriptor.

## What crosses

Nothing in memory survives an exec, so whatever the new image needs is written
down first and handed over by descriptor number:

- **The PTY masters**, duplicated so the old actor can drop its own copy normally.
- **The replay rings**, so the screen is rebuilt rather than redrawn.
- **The already-bound Unix listener.** Re-binding would mean unlinking and
  re-creating the socket, and a client connecting in that window would autostart
  a second daemon over a state directory this one still owns. Carried, connects
  queue in the kernel backlog instead.

The blob is written to a file that is unlinked before a byte goes into it: it
holds raw terminal output, and a handoff must not become a way of writing that
to disk behind the user's back.

What does not cross: client connections (they reconnect and reattach by session
id, to find the session alive), the WebSocket listener, and the VT sidecar.

## Descriptor ownership is the invariant

Everything handed over stays *owned* until the crossing is certain. A session's
master is an `OwnedFd` from the moment the actor duplicates it, and `pack`
releases ownership only after the last fallible step has succeeded. Two things
fall out of that, and both were bugs before it:

- A session that misses the carry deadline has its master **closed**, so it is
  hung up like any other loss — not left running, un-hung-up, inside a session
  the new daemon has no handle for.
- A write failure partway through packing drops every master rather than
  scattering raw descriptors nothing owns.

## Order, and what it does not buy

Everything fallible that does not need the sessions runs first: resolving the
state directory, creating and unlinking the blob file, duplicating the listener.
Only then are the actors asked for their PTYs.

That does not make the rest safe, and the code no longer claims it does. Writing
the blob can still fail, and the new image can still fail to start. The far side
answers that by degrading rather than exiting — the runtime directory, the
WebSocket configuration and the tombstone log all log and carry on, because on
this path refusing to start means killing what you were just handed. The host id
is carried in the blob rather than re-read, which removes that failure outright
and states that identity does not change across a handoff.

## Vetting the candidate

`--version` is not the question. Every termiod ever built answers it, including
the ones that have never heard of `serve --handoff`; exec one of those and the
daemon is replaced by a program that exits. The candidate now has to answer
`handoff --probe` with the exact format token, on a five-second deadline with a
capped reply, so a candidate that hangs or floods cannot wedge the verb.

**This is still self-attestation.** A binary that answers correctly and then
fails to start takes every session with it, and nothing short of running
`serve --handoff` for real would prove otherwise — which is the step that cannot
be taken back. The residual is written down in the module docs rather than
implied away. What it closes is the case that actually happens: a staged binary
too old to know the verb.

## Screens

A carried session whose ring can no longer draw its own screen — it evicted
output, or was resized after those bytes were written — comes back with its VT
stale, so snapshots fall back to ring replay instead of the host asserting a
grid it reconstructed from bytes it has been told are wrong.

## Deploy

`deploy` reaches for `handoff` first and falls back to `stop` for a daemon too
old to advertise the capability. A busy box is now upgraded rather than staged.
An explicit `--binary` is verified against that binary's own version, so a
deliberate rollback is not reported as a failure.

## Verification

191 unit tests and 6 integration tests in `tests/handoff.rs`. The load-bearing
assertion is a counter file: a shell writing to it every tenth of a second stops
the instant its PTY master closes, so a counter still moving after the daemon has
replaced its own image is the feature, observed from outside. The rest cover the
convincing impostor, the hanging and flooding probes, the listed-or-gone
invariant, and — against the session actor, where a real client holds the write
token — the resize that makes a ring stop describing its screen.

Two rounds of adversarial review found eleven defects between them; the commits
say which. `tests/join_invariant.rs` fails on this branch and on `main` alike —
pre-existing, not touched here.

Release Notes:

- termiod can now replace its own binary in place. Upgrading a machine keeps its
  sessions, their scrollback and everything running in them; clients reconnect.

### DIFF
--- a/termiod/DEPLOY.md
+++ b/termiod/DEPLOY.md
@@ -22,10 +22,9 @@ export DEVELOPER_DIR=/Library/Developer/CommandLineTools
 cargo build
 alias tio=./target/debug/termiod
 
-# 2. Deploy to the VPS. Two gotchas: a RUNNING daemon keeps executing the old
-#    binary image (and scp onto a running binary fails with ETXTBSY), so stop
-#    it first. Existing sessions die with it — this is a dev-loop step.
-ssh ukvps pkill -x termiod || true
+# 2. Deploy to the VPS. The running daemon keeps executing the old image until
+#    it is told otherwise, so deploy stages the new binary beside it and then
+#    asks it to hand off — same pid, same PTYs, sessions intact.
 tio remote deploy ukvps
 
 # 3. Open a durable session and attach (creates + attaches in one command;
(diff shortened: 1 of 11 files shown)


jiweiyuan:
Adversarial review found three blockers, all in the loss mode this feature exists to prevent: (1) critical — any post-carry `pack` failure (ENOSPC, sync, seek) drops every carried master with no rollback, killing all sessions at once (daemon.rs:736/761/662); (2) high — `Control::Send` resolves a stale handle in the carry-to-exec window, ignores `send()`'s `false`, and returns `Ok`, so accepted input is silently discarded (session.rs:1625, daemon.rs:1250); (3) high — a healthy session missing the 2s `Carry` deadline is sacrificed rather than aborting the upgrade (daemon.rs:106, session.rs:1633). Descriptor mechanics, listener inheritance, and tombstone retention all reviewed sound.

PR #523, opened 2026-08-29
Supervise termiod with a systemd --user unit on Linux

jiweiyuan:
Making termiod the only PTY owner makes it a single point of failure. macOS has had `termiod service install` (a launchd agent) for that; Linux bailed with "macOS-only" and pointed at a hand-written unit, so a VPS reboot or an SSH logout could take every session with it.

## What changes

- `termiod service install | uninstall | status` work on Linux with the same clap surface. `install` writes `~/.config/systemd/user/termiod.service` (`termiod-dev.service` on the dev channel) with an absolute `ExecStart`, `Restart=on-failure`, `WantedBy=default.target`, runs `enable --now`, and tries `loginctl enable-linger` — printing `loginctl enable-linger $USER` as the next step when polkit refuses it over SSH. `TERMIO_CHANNEL` is pinned off the release channel and `TERMIOD_SOCK` only if the caller had it set, the same rules as the plist. No `--wss` in the unit; the bind survives in `wss.bind` or a drop-in.
- `uninstall` disables and removes the unit and leaves linger as it was; it says whether sessions were stopped. `status` is dense like the launchd one: unit file, socket, `active (pid N)` / `inactive (enabled)` / `not installed`, linger.
- **No dual start.** When a unit owns the caller's socket, `spawn_daemon` runs `systemctl --user start` instead of forking a `setsid` daemon. After a clean `termiod stop` the unit is merely inactive (`Restart=on-failure` does not restart a clean exit), and a fork would take the socket and leave the unit failing at every later start — demoting the box to unsupervised without anyone asking. The check reads the unit file first, so an unsupervised box spawns no process, and a `TERMIOD_SOCK`-pinned caller (tests, a side-by-side daemon) never starts the real unit.
- `detect_supervisor` uses that same check and the channel-scoped name; it hardcoded `termiod` before. `deploy` keeps SIGTERM as its stop; the bounce path is documented at the stop step.
- `install` refuses while an unsupervised daemon holds the socket, since the unit's daemon would exit "already running" until the start limit tripped.
- Metacharacters in the binary path cannot break the unit (`%`, `$`, `"`, `\`, space, newline are quoted/escaped).

## Docs

`service.rs` module header, `DEPLOY.md` (the hand-written unit becomes `termiod service install`; the WSS section uses a `systemctl --user edit` drop-in), and one sentence each in the lifecycle RFC §4 and §11.

## Verification

- `cargo build` and `cargo build --target aarch64-unknown-linux-musl`: clean (the 8 warnings are pre-existing, in untouched files).
- `cargo test --bin termiod -- service:: lifecycle::`: 20 passed, including 5 new unit tests mirroring the plist ones.
- Exercised on a real box (Ubuntu aarch64, `systemd --user` running, linger already on) with this PR's musl build on the dev channel, so the release daemon and its sessions were never touched:
  - `service install` wrote `termiod-dev.service` with the absolute `ExecStart`, `Restart=on-failure` and `Environment="TERMIO_CHANNEL=dev"`; `status` reported `active (pid 3194178)`, `linger: yes`; `termiod status` reported `service: systemd --user`.
  - `termiod stop` → `systemctl --user is-active` = `inactive`, `is-enabled` = `enabled`; `status` still says `systemd --user` with no daemon running.
  - `termiod list` on the dead socket brought the daemon back with its pid equal to the unit's `MainPID` (3194516) — started through `systemctl --user start`, not a `setsid` fork.
  - `service uninstall` removed the unit and reported the stop; a following `list` autostarted an unsupervised daemon, and `service install` then refused with "a daemon is already running unsupervised on …; run `termiod stop` first" (exit 1). After `stop`, install and uninstall ran clean and `status` reported `not installed`.

Release Notes:

- `termiod service install` now supervises the daemon on Linux with a systemd `--user` unit, so sessions survive a crash, a logout (with linger), and a reboot.

### DIFF
--- a/docs/design/20260827-termiod-lifecycle-reconcile.md
+++ b/docs/design/20260827-termiod-lifecycle-reconcile.md
@@ -3,7 +3,7 @@ title: termiod lifecycle — install and update as one reconcile loop
 status: in-review
 type: rfc
 created: 2026-08-27
-updated: 2026-08-27
+updated: 2026-08-29
 related:
   - 20260825-agent-integration-moves-to-termiod.md
   - 20260824-agent-integration-on-a-device.md
@@ -144,8 +144,8 @@ stays as a developer fallback and is never reached from the app.
 
 **The daemon is handed to the OS supervisor when there is one.** `service.rs`
 does this for launchd. The Linux half — a systemd `--user` unit plus
-`loginctl enable-linger` — is specified there and not built. This RFC does not
-build it either; it records that `status.supervisor` is the field that will
+`loginctl enable-linger` — is specified there and was built after this RFC
+(2026-08-29); it records that `status.supervisor` is the field that will
 tell the loop whether "restart" means `systemctl --user restart termiod` or
 "stop it and let the next client autostart it". Both paths converge on the same
 loop; only the activate step differs.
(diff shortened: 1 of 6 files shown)


jiweiyuan:
Superseded: the same change landed on main as a1a8cdc via #522 (live-tested on a systemd box, results in the description above).

PR #535, opened 2026-08-30
Stop the reattach repaint eating the last line on screen

jiweiyuan:
Quit the app while a session streams output, reopen it, and the line written in the attach second vanished from the screen — reproducibly, every attach. The daemon's authoritative VT always had the line (an observer attach showed the full sequence with zero gaps); only the reattached app's screen lost it.

The mechanism, taken from a byte tap between the app and the daemon: the snapshot formatter serialises a scrolled primary screen as one newline-joined flow that stops at the last non-blank row. The blank rows beneath it — the scroll steps the host has already taken — never reach the client, so the replayed screen sits one row behind the host's. The trailing cursor re-assert then lands *on* the last painted line instead of the blank row under it, and the first live byte after the seam overwrites that line in place.

The fix stays in `termiod-vt`'s `format_vt`, and the engine itself referees: replay the payload into a scratch terminal of the same grid, and while its screen trails the live one, feed it the newlines the flow dropped — adopting the padding only once the replay converges on the live screen. Payloads the formatter already positions correctly (alt-screen frames, unscrolled screens, pending-wrap restores) pass through untouched, and the shortfall search is bounded so a pathological screen can't turn a snapshot into a grid crawl. The active-screen reads go through `grid_ref` like history cells, so damage tracking is never touched.

Verified three ways:
- `termiod/vt/tests/attach_replay.rs`: five replay tests, including the captured bug shape; the two scroll cases fail without the fix and pass with it.
- `cargo test` across the termiod workspace: all green.
- Live: with the fixed daemon, five quit-relaunch rounds against a streaming counter session reattached with zero lost lines (the same experiment reproduced the loss on every round before the fix).

Release Notes:
- Fixed: reopening the app no longer loses the line a session printed at the moment of reattach.

### DIFF
--- a/termiod/vt/src/lib.rs
+++ b/termiod/vt/src/lib.rs
@@ -251,6 +251,35 @@ impl VtTerminal {
             formatted.extend_from_slice(&bytes);
         }
 
+        let pending_wrap = check(
+            self.terminal.is_cursor_pending_wrap(),
+            "Terminal::is_cursor_pending_wrap",
+        )?;
+
+        // The formatter paints a scrolled primary screen as one newline-joined
+        // flow and stops at the last non-blank row, dropping the blank rows
+        // beneath it — and with them the scroll steps the host has already
+        // taken. A client replaying the payload then holds a screen sitting a
+        // row (or more) behind the host's, so the cursor re-assert below lands
+        // *on* the last painted line instead of the blank row under it, and the
+        // first live byte after the snapshot overwrites that line in place.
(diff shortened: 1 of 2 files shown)


jiweiyuan:
Review found the 8-step cap left 9+ newline shortfalls unrepaired; 3892dff derives the bound from the live screen's trailing blank run instead, with regression tests at 12 scrolls and a fully blank screen.

PR #548, opened 2026-08-31
Keep the Mac awake while a session is busy

jiweiyuan:
While an agent streams, its pty counts as an active tty and pmset's default `ttyskeepawake` keeps the machine up for free. The moment it prints a permission prompt and goes quiet, that cover lapses, the idle-sleep timer runs out (`sleep 1` is a common default — one minute after display sleep), and the phone that was supposed to answer the prompt can no longer reach the box. The agent that most needs the machine reachable is exactly the one that can't keep it awake.

## What this does

`termiod` now renews a short `caffeinate -i -t 120` lease every 60 seconds while any session is busy — the same `SessionSummary::busy()` predicate the update path uses (`running || working || needs_you`), protocol status rather than attachment. The first lease lands on the status event itself, not the next tick.

The assertion is a lease, not a held lock. Nothing is ever released: a crash, `termiod stop`, and the `execve` handoff (which runs no destructors and would orphan any long-lived child) all leave at most one lease that expires on its own. The lease tail doubles as the ~2-minute grace period after the last session goes quiet.

Policy edges:

- **Battery**: no lease is taken at all. This has to be the daemon's own `pmset -g batt` check — `caffeinate -i` is honored on battery, and `-s` only adds a second AC-only assertion rather than scoping `-i`.
- **Display**: sleeps either way; `-i` touches idle *system* sleep only.
- **Lid**: a user-space assertion cannot stop clamshell sleep. Out of scope.
- **Linux**: never spawned — the boxes termiod serves there don't idle-sleep.
- **Opt-out**: `serve --no-keep-awake`, or `TERMIOD_KEEP_AWAKE=off` in the daemon's environment. The flag rides through the handoff via `serve_arguments()`.

Integration tests set `TERMIOD_KEEP_AWAKE=off` on every long-lived `serve` they spawn, so test daemons never take real power assertions on developer machines or CI.

## Verification

- `cargo test`: all green, including 4 new unit tests on the renewal decision (a burst of status events costs one lease, not one per event).
- Live smoke on an isolated channel: `working` produced a `caffeinate -i -t 120` child of the daemon within 3s; 70s after `done`, the same (expiring) lease and no renewal; `stop` left nothing behind.

Release Notes:

- The Mac now stays reachable while an agent is working or waiting on you: termiod holds a short idle-sleep assertion (AC power only, display still sleeps) and lets it lapse a couple of minutes after the last session goes quiet. Opt out with `termiod serve --no-keep-awake` or `TERMIOD_KEEP_AWAKE=off`.

### DIFF
--- a/termiod/src/daemon.rs
+++ b/termiod/src/daemon.rs
@@ -313,6 +313,10 @@ impl Manager {
         let _ = self.events.send(event);
     }
 
+    pub(crate) fn subscribe_events(&self) -> broadcast::Receiver<Event> {
+        self.events.subscribe()
+    }
+
     async fn info(&self, handle: &SessionHandle) -> Option<SessionInfo> {
         let (tx, rx) = oneshot::channel();
         if handle.send(SessionMsg::Info { reply: tx }) {
@@ -347,7 +351,7 @@ impl Manager {
         });
     }
 
-    async fn list(&self) -> Vec<SessionInfo> {
+    pub(crate) async fn list(&self) -> Vec<SessionInfo> {
         let handles = self.handles();
         let mut infos = Vec::new();
         for handle in handles {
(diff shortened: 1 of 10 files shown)


jiweiyuan:
Codex review: fixed execve zombie leases (leases now detach to launchd) and made a failed lease spawn retry instead of disabling keep-awake.

PR #549, opened 2026-08-31
Make the pair CLI adopt and revoke legacy state

jiweiyuan:
Closes out the review findings on #534 (merged): three `termiod pair` paths still bypassed the durable-state split's adoption.

- `--rotate` wrote the new durable token but left the legacy copy beside the socket — a live credential a pre-split daemon keeps honouring per handshake, while the command reports every client signed out. Rotation now removes it as part of the revocation.
- `--wss-off` removed only the durable bind; the next daemon start adopted the legacy one and re-armed the listener the operator had turned off. Both copies now go through one `paths::remove_wss_bind` helper.
- `--json`/`--qr` read `wss.origin` without adoption and refused an invite whose origin only existed beside the socket. `run_pair` now adopts up front.

Verification: three new integration tests in `termiod/tests/pair_state.rs` run the real binary in a sandboxed `HOME` + `XDG_RUNTIME_DIR`; all three fail against the unpatched sources and pass with the fix. Full termiod suite: 249 passed, 0 failed.

Release Notes:
- Fixed: `termiod pair --rotate` now revokes the pre-upgrade pairing token everywhere, `--wss-off` stays off across daemon restarts, and invites work before the upgraded daemon's first restart.

### DIFF
--- a/termiod/src/paths.rs
+++ b/termiod/src/paths.rs
@@ -422,7 +422,40 @@ pub fn load_or_create_pair_token() -> Result<String> {
 pub fn rotate_pair_token() -> Result<String> {
     let token = encode_base64url(&random_bytes(24)?);
     replace_secret(&pair_token_path()?, &token)?;
+    // Rotation is revocation, and the legacy copy beside the socket is still a
+    // live credential: a pre-split daemon reads it per handshake, and it would
+    // keep admitting the old secret until a restart while `--rotate` reports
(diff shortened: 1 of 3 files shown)


jiweiyuan:
Review found --wss-off could race an in-flight startup adoption and resurrect wss.bind; 60f91c7 serializes both on a flock in the durable dir, with a 12-round race test.

PR #553, opened 2026-08-31
Extract termiod's module tree into a library crate

jiweiyuan:
The crate had one binary rooted at `src/main.rs` declaring all twenty modules, so a second binary could not import them. This moves the module tree to `src/lib.rs` (every module `pub`; the crate is internal and marked `publish = false`) and makes `src/bin/termiod.rs` a thin entry importing what it uses. `Metered` and `ClientBacklog` become `pub` because public enums in a library may not expose crate-private types.

This is the lib extraction the docker-lessons RFC §1.3 prices as the prerequisite for the Rust `termio` client (unify-server-plane Stage 10, pulled forward per #551). The `termio` binary itself comes in a follow-up; this PR is deliberately mechanical.

Verified no behavior change:

- `--help` output byte-identical to the pre-move build, top level and all eighteen subcommands
- `cargo test` green: 255 tests across the lib and 11 integration suites, including the 8 handoff-contract tests
- `smoke_test.py` and `remote_smoke_test.py` both pass
- `--version` output identical; CI's `TERMIO_TERMIOD_TEST_BIN` path (`target/debug/termiod`) unchanged

Release Notes:

- None (internal refactor).

### DIFF
--- a/termiod/Cargo.toml
+++ b/termiod/Cargo.toml
@@ -2,6 +2,7 @@
 name = "termiod"
 version = "0.1.0"
 edition = "2021"
+publish = false
 description = "termiod — durable PTY session host (POC). Detach ≠ kill."
 license = "MIT"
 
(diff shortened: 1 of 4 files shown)


jiweiyuan:
Adversarial review found no runtime regression (build.rs, CI, doctests, statics all checked); its executable-identity findings are now stated on the crate root in ccd0d03 and respected by #554.

PR #556, opened 2026-08-31
Port sessions, notify, and agent report to the termio client

jiweiyuan:
The parity slice of the Rust `termio` client (unify-server-plane Stage 10, following #551/#553/#555): every `sessions` verb, `notify`, and `agent report` move into the binary, still speaking the app's control socket. Backends change one verb at a time in later PRs, each flip retiring its Swift handler; this PR is the byte-compat heavy-lifting that makes those flips diffable.

- **`app_socket.rs`** reproduces the shell client's request shape byte for byte — field order, its exact `json_escape` semantics (C0 stripping, ESC surviving as a JSON u001b escape), the `is_address` heuristic — plus its error taxonomy (sandbox denial / missing socket / stale socket / silent app), the `--wait` exit-code contract (0/1/3), and watch's stream semantics (0 on Ctrl-C, 1 on an error first line, 2 on EOF with the escalation message).
- Two load-bearing macOS traps found while porting, documented in place: the app treats a half-closed request socket as the client leaving (`nc` never shuts down its write side, so neither do we), and **setsockopt is refused with EINVAL once the peer has closed** — both timeouts are armed before the request goes out and never re-armed, or every fast reply becomes a phantom "connection refused".
- `$PWD` is validated against the physical cwd the way `sh` validates it before being trusted as `caller_cwd`.
- `agent report` execs the located daemon's `set-status` with the script's exact argv and keeps both silent-success shapes (`{}` under `--reply`).
- **`tests/cli_compat.py`**: both clients run against a mock app socket; request bytes, stdout, stderr, and exit codes are diffed across **76 cases** — every verb in both formats, `--wait` settle/timeout, watch stream/error/EOF, all validation errors, the error taxonomy, the hook contract (argv-recording fake daemon), and every help text. Wired into the macOS CI job (Linux would diff the script's own `nc` platform drift, not the port).

Verified: all 76 harness checks pass; 267 tests green across the lib and integration suites; `list`/`read`/`focus` additionally diffed byte-identical against the live app in both formats, including error paths.

Release Notes:

- None (the binary is not yet bundled; it ships to users in a later PR).

### DIFF
--- a/.github/workflows/termiod.yml
+++ b/.github/workflows/termiod.yml
@@ -138,6 +138,14 @@ jobs:
           pkill -9 -f "target/debug/[t]ermiod|deps/[t]ermiod-" || true
           python3 remote_smoke_test.py
 
+      # The Rust termio client against the shell client it replaces: request
+      # bytes, stdout, stderr, and exit codes must be identical on every verb
+      # (unify-server-plane Stage 10's freeze). macOS only — the shell client
+      # itself depends on macOS nc semantics, so a Linux run would diff the
+      # script's own platform drift, not the port.
+      - name: CLI compat harness
+        run: python3 tests/cli_compat.py
+
       # The Swift client's half of the files plane. These tests spawn a real
       # daemon on a private socket, so `swift test` skips them unless
       # TERMIO_TERMIOD_TEST_BIN names a binary — and the Swift workflow's runner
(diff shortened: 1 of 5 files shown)


jiweiyuan:
Adversarial review found five real parity breaks (timeout sign, trailing-newline escape, close word-splitting, report argv word-splitting, watch unterminated lines) — all fixed in 4043da6 with regression cases; the harness now runs 91 checks including the parsing-order contract.

PR #557, opened 2026-08-31
Bundle the Rust termio client in place of the shell script

jiweiyuan:
PR 4 of the CLI port (after #556): `build-app.sh` lipos the `termio` bin out of the same cargo slices the daemon build already produces — the client costs no second build — and ships it under the channel's resource name (`termio`/`termio-dev`), sealed with the hardened runtime like the daemon so notarization holds. The sed-rewritten shell client survives only as the dev-without-Rust-toolchain fallback. `CommandLineTool` needed zero changes: the support copy and PATH symlink flow treats a Mach-O like any content, so hooks written by old releases keep resolving on the exact same path.

Two latent bugs became load-bearing and are fixed here:

- **Stale support copy after update**: the copy only refreshed when the hooks toggle changed or on refocus with hooks on — an updated app could leave an old client on the exact path hooks exec. `TermioStore` now refreshes it unconditionally at launch.
- **Channel rebinding via inherited env**: `AppChannel.suffix` let an inherited `TERMIO_CHANNEL` beat the bundle's own identity. Every termio session carries its app's channel in that variable, and macOS `open` propagates the caller's environment — so a dev app launched from inside a release session silently rebound itself to the *release* channel's state directory and sockets. The app's own bundle id now wins for the two real channel names; custom probe channels stay honoured, and unbundled `swift run` keeps the env override.

Also: the client restores default SIGPIPE disposition, dying silently on a closed pipe (`termio version | head`) like the script instead of panicking mid-print.

Verified: dev `.app` built and launched — the support copy becomes the signed Mach-O with 0755 at launch; the installed copy drives the dev app (`sessions list` prints `termio-dev://` links) and passes all 91 compat-harness checks; release support copy untouched; 225 lib tests, integration suites, and `swift test` green.

Release Notes:

- The `termio` command-line tool is now a native binary. Its commands, output, and paths are unchanged.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -54,6 +54,7 @@ swift test                                 # run the unit tests
 ./scripts/build-app.sh                     # ad-hoc-signed .app → ./termio.app
 TERMIO_CHANNEL=dev ./scripts/build-app.sh  # side-by-side dev app → ./termio-dev.app
 ./ios/dev-run.sh                           # build + install iOS app on the device
+cargo build --bin termio                   # (in termiod/) the Rust CLI, for checkout use
 ```
 
 termio is a real foreground AppKit app bootstrapped by an explicit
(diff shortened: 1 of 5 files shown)


jiweiyuan:
Adversarial review found no defects across the AppChannel precedence, signing order, fallback path, refresh idempotence, and SIGPIPE checks — mergeable as is.

PR #560, opened 2026-08-31
Serve sessions read from the daemon

jiweiyuan:
PR 5 of the CLI port — the first backend flip (unify-server-plane Stage 10, one-path Stage 6 verb 1). A `sessions read` whose target the local daemon owns is answered from the daemon's authoritative VT: an observe-mode attachment negotiated with the `snapshot` capability, dropped once the `S` frame lands. Observe mode takes no write token and never resizes, so a read is invisible to the session's real client. The v2 payload replays into a fresh `termiod-vt` engine and renders rows the way the app renders a screen (right-trimmed, trailing blanks dropped, `--lines` tail, same sorted-key `--json` shape).

This is the verb's designed behavior change, and it showed on a live session during verification: the app path answers `not_live` ("open it once in Termio") for a session no window ever opened, while the daemon path prints its actual screen. The verbs also work on a box with no Mac app at all — the Linux CI job now proves it.

The router keeps an app-socket fallback for coverage the local daemon lacks — sessions on remote devices (their PTYs live in the far daemon; the app's surface is the only local view), app-side title matching, foreign-channel links — so the Swift `readScreen` handler survives until the CLI can route to remote daemons. The new daemon connections never auto-start a daemon: one this call would have to start hosts no sessions, and from this binary `current_exe()` is not the daemon (the `lib.rs` seam).

Tests: `tests/cli_read_smoke.py` pins a private daemon and proves the gates (read by prefix and link with no app socket anywhere, JSON shape, `--lines`, unowned-target fall-through); it runs in both CI jobs. The compat harness pins a dead `TERMIOD_SOCK` so its app-socket diffs stay deterministic — all 91 cases still pass, and 274 crate tests are green.

Release Notes:

- `termio sessions read` now answers from the session host, so it works for sessions never opened in a window — and on Linux boxes with no app.

### DIFF
--- a/.github/workflows/termiod.yml
+++ b/.github/workflows/termiod.yml
@@ -138,6 +138,13 @@ jobs:
           pkill -9 -f "target/debug/[t]ermiod|deps/[t]ermiod-" || true
           python3 remote_smoke_test.py
 
+      # The daemon half of `sessions read` against a real daemon — the Stage 10
+      # gate that a never-opened session answers, with no app socket anywhere.
+      - name: Read smoke test
+        run: |
+          pkill -9 -f "target/debug/[t]ermiod|deps/[t]ermiod-" || true
+          python3 tests/cli_read_smoke.py
+
       # The Rust termio client against the shell client it replaces: request
       # bytes, stdout, stderr, and exit codes must be identical on every verb
       # (unify-server-plane Stage 10's freeze). macOS only — the shell client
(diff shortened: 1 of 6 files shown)


jiweiyuan:
Both blockers and both smaller findings fixed in the follow-up commit: stale-VT reads rebuild from ring replay under a 10s deadline, daemon-first matching is restricted to UUID-shaped names, link parsing mirrors addressedID with tests, and the smoke test polls instead of sleeping.

PR #568, opened 2026-08-31
Derive agent status on the device, and ladder the companion wire's retirement

jiweiyuan:
The phone speaks a second protocol to the Mac. That makes the Mac a server as well as a client, and routes a session running on a VPS through a laptop at home — two VT parses, a geographic detour, and a Mac that must be awake for work that is not on it. Invariant #4 says one protocol, versioned and transport-agnostic. This is the one place the repo still says otherwise.

Two things here: the RFC that retires it, and its first stage.

## The RFC

`docs/design/20260831-companion-second-protocol-retires.md`. Eight stages, each shippable alone with a gate that runs, ordered by one rule: **no stage may leave the phone worse than the relay it replaces.**

It re-decides nothing the three RFCs that own pieces of this already settled — the listener is the web-client RFC's, the iOS seam is `20260824-ios-as-device-client.md`'s, the invariants are device architecture's. That last one stops exactly where this starts: its own non-goals say the companion wire *"is retired when the phone and the browser both attach directly"*. It made a second path possible; it did not say what happens to the state the first path was carrying. Every remaining question is of that shape — not transport, **ownership**.

So this decides three things: where each piece of phone-visible state lives once the Mac is a peer (§3, §4), the order the pieces move in (§5), and the four planes no other document had placed — notifications, the permission broker, transfer, and Usage (§6).

It also answers the three questions unify-server-plane Stage 6 explicitly left for whoever came next, rather than letting code review decide them:

- **The attach replay bound** — no wire change. `RING_CAP` is already 128 KiB, the exact number Stage 6 named as what the phone needs, and a snapshot-negotiating client gets `S` at the boundary with no ring replay at all. The question was written against `PTYProcess`, whose ring was the whole scrollback.
- **`jiggleResize`** — deleted. A host-side redraw op is the host deciding what a client should be looking at. `request_snapshot` already shipped for the same need in #517.
- **Ownership reconciliation** — settled by #517, in the daemon's favour: the token moves by typing only. One stale comment survives.

## Stage 1 — agent status moves to the device

This reverses `20260819-unify-server-plane.md` §4.1, which lists the status rules under *Stays Swift* and tells a future stage to argue against the list rather than around it. §3 argues.

The list's reasoning was that every client receives the same bytes and already holds the rendered viewport. Both true; neither sufficient. **The rules are not functions of the bytes.** They also read this client's scroll position, this client's keystrokes, this client's tick history, and this client's selection — so two viewers running them independently disagree by construction. And the asymmetry is worse than that: a phone attached straight to a device today gets `E status` and nothing else. No title channel, no progress channel, no screen rules, no stall probe, no promotion. For an agent with no hook system that is *no status at all* — on a home screen whose organising idea is a **Needs You** strip. Deferring this makes direct attach a downgrade, which is why it is Stage 1 and not a cleanup.

**What moved** into `termiod/src/session/status.rs`, beside the sidecar VT and off the byte path: declared screen-rule classification, `OSC 0/2` title classification, `OSC 9;4` progress, the two-tick streak promotion with all four of its guards, the stale-working sweep, and the four-probe stall detector.

**What did not**, because it is presentation: the tooltip sentence, argv → glyph, and the done-versus-idle call. That last one is the interesting seam — a `done` the agent itself reported means `done` on every client, while a turn the *device* concluded is a question about whether this person has looked yet. `E status` grows a `source` field so a viewer can tell the two apart, and applies its own focus to the second.

Three defects go away as a consequence, not by being fixed:

- `readViewportText()` returns the viewport a user has scrolled, so scrolling a pane fed stale rows to the classifier. The sidecar VT has no scroll position.
- `noteUserInput` existed so keystroke echo would not read as agent activity, and was fed from the Mac's write path. The daemon's PTY write path is the real choke point — it sees the phone and `termio sessions send` too.
- One engine, one verdict.

**Shape:** the `status:` resource — one per device, `seq` + bounded ring + replay-or-gap + linger, the same mechanism `fs:` and `git:` already ride. A phone that locks mid-turn resumes at its cursor. `E status` stays for the attached-client case; both come out of the one engine.

**The recorded objection, answered.** `termiod/src/agent/fixture.rs` argued that choosing a Rust regex engine means choosing a different accepted language — *"a worse kind of disagreement than none."* That is right while both sides match, and stops applying when only one does. The narrowing is real (`regex` is RE2-shaped: no lookaround, no backreferences), bounded, and now asserted: every pattern termio ships compiles, as a test. A user's own pattern that does not is dropped with a log, which is what Swift already did with an invalid one.

The Swift engine is **deleted, not disabled** — no dual engines, no flag. `AgentStatusRules` survives as pattern source with no matcher, because the app still parses manifests to render the roster and the fixture contract says both parsers stay.

`termio agent report` is untouched — every flag, the silent no-op outside a session, the `{}` on `--reply`, and the daemon-resolved binary path that keeps global config from ever naming something inside `Bundle.main`.

## Verified

- `cargo test` — 320 passed, 0 failed. 32 of those are the Swift cases ported case for case: `StallProbeTests`, `AgentTitleStatusTests`, `OSCProgressScannerTests`, plus the arbitration precedence rules and the two new `status:` resource cases.
- `swift build` clean, `swift test` — 932 passed, 0 failures.
- `xcodebuild` for the iOS simulator — clean.
- The engine is gone: `grep -rn 'firstMatch\|NSRegularExpression' Sources/termio/Agents Sources/termio/TermioStore` → 0.

**Not verified on a device pair.** Stage 1 changes what the device says, not how a phone reaches it. The phone's direct backend already subscribes to `status`, so it gains the three derived channels the day it attaches to a box running this — but that is a build-time claim until Stage 2 gives it a device to attach to.

## Why one PR

The §4.1 reversal is not credible as prose alone. The argument is that identical bytes do not produce identical verdicts, and the thing that shows it is the ported test suite sitting next to the deleted one.

Release Notes:
- Agent status now comes from the machine the agent runs on. A session on a remote box reports the same working / waiting / done as one on your Mac, including for agents that have no hooks and are read off their screen — and every device you watch it from shows the same thing.


jiweiyuan:
Review round 1 — four of five fixed, one disputed with evidence. `05b5eea`.

**1. `status:` had no subscriber — real, fixed.** Neither client called `subscribe_resource("status:")`, so the registry never built a ring and `publish_status` returned early on every batch: an advertised seq/replay/gap mechanism that had never once run. I chose to wire the consumer rather than defer the resource, because the phone is the client the cursor exists for.

`TermiodBackend` now subscribes to `status:` with its last cursor **in place of** the `status` event name, so there is one path and not two that can drift. `stalled` already rode those batches, so dropping the broadcast loses nothing. `deviceCapabilities` gains `resources`. A device too old to know the resource answers `error`, and the phone falls back to the broadcast rather than going dark. On `gap` it drops its deltas and lets the `list` already in flight re-establish them.

The Mac stays on `E status` deliberately: it attaches to every session it shows, so it already has the per-session channel. A roster-wide cursor is for a client watching sessions it is *not* attached to, and on the Mac that is Stage 3.

**2. `source` / `turn_ended` dropped on iOS — real, fixed.** Correctly identified as a disagreement this PR introduced. The phone now applies the same rule from its own open session (`viewingSessionID`, forwarded from `RosterStore.currentSessionKey`, which is the device session id for a live row), and opening a session clears the mark the way looking at a row does on the Mac.

Six new iOS unit tests, driven through the real decode and the real `publishRoster`. The first cut of them passed on an empty backend and proved nothing — they now seed a roster through the `sessions` reply and a `hello_ok`, and read the status off the published roster the lists actually render from.

**3. Handoff reset the clocks — real, fixed.** `CarriedSession` carried a status string but not how long it had been true, so adoption restarted both the stale-working sweep and the 20-minute stall window. It now carries elapsed durations (`status_clocks`), additive with a serde default so an older blob still crosses — `a_carried_session_without_status_clocks_still_decodes` pins that.

The stall **baseline** deliberately does not cross: this process cannot vouch for a repo it did not watch, so the next sweep re-captures. The window *start* does cross, and a capture does not move it, so an already-elapsed window still probes on its next tick.

Your suggested test is in, plus the window/latch case:
- `a_handoff_keeps_the_stale_working_clock_on_its_original_schedule` — quiet 11s, hand off, assert the sweep fires 1s later and not 12s later.
- `a_handoff_keeps_the_stall_window_and_its_latch`
- `an_older_blob_with_no_clocks_still_adopts`
- `a_quiet_session_carries_nothing`

**4. `request_snapshot` has no client caller — this one is not so.** Both clients send it, and have since #517:
- Mac: `TermiodClient.requestResyncLocked` → `Termiod.requestSnapshotPayload()` (`TermiodClient.swift:1475`), armed by `observerRepaintPending`.
- iOS: `TermiodSession.swift:143`, same flag, same payload.

What I *did* overstate is the other half. §7.2 said "jiggleResize is a client-side wipe plus `request_snapshot`", and nothing implements a wipe. It now splits the two: the ask shipped and is wired end to end; the wipe does not exist and nothing needs it, because a snapshot is a full repaint from a known boundary and `jiggleResize`'s only caller went with `PTYBridge` in Stage 6. If a case ever needs one it is a client-side clear plus the verb that already exists — never a host-side redraw op.

**5. Gate record inaccurate — real, fixed both ways.**

The numbers: `329 passed, 0 failed` across all binaries — **274 lib** (40 of them `session::status`), 3 bin, 52 integration. The old "320" was a cross-binary sum labelled ambiguously; the RFC now breaks it out.

The fidelity claim: you were right that the OSC port consolidated. 14 Swift cases became 5. The RFC now carries a case-by-case table, and the four genuinely missing ones are restored as Rust tests:
- `every_keepalive_is_reported_within_a_chunk_and_across_them`
- `an_overlong_payload_is_rejected_rather_than_truncated` (bound moved 24 → 1024 because titles ride the same scanner, so it also asserts the scanner recovers)
- `a_report_amidst_other_output_still_lands`
- `classify_pins_the_grammar`

That surfaced a real defect in the port: the scanner's doc comment claimed it collapses consecutive duplicates within a chunk, and it does not. The behaviour is right — the arbiter collapses them, and it refreshes the liveness clock *before* that guard, so each keepalive stays evidence — but the comment was wrong and is fixed. Two merges are deliberate and named in the table; one case (`testTitleAndNotificationAndPaletteAreIgnored`) narrowed on purpose, because a title is no longer ignored — it is read as a title and never as progress.

## Verified

- `cargo test` — 329 passed, 0 failed
- `swift build` clean, `swift test` — 932 passed, 0 failed
- `python3 termiod/tests/cli_compat.py` — **93/93**
- iOS unit tests — 34 passed, 0 failed (8 new)
- iOS simulator build — clean

One note on the cli_compat hold: it read 28/65 on the first check of this round. That was a stale `target/debug/termio` — this worktree symlinks its `target/` to the main checkout, and a build elsewhere had clobbered the binary with a pre-#559 one still naming `session-control.sock`. Rebuilt and it is 93/93. Not a regression, but worth recording since the symlink will do it again.

Two iOS **UI** tests fail (`testInspectorFileTreeShowsLanguageIcons`, `testScanEntryPresentsScanner`). I ran both on `origin/main` with this branch stashed and they fail identically there; neither touches status. Recorded in the RFC rather than folded into the count.

Still not verified on a device pair, and the RFC now says so plainly: the phone's subscription, cursor and focus rule are unit-tested against the daemon's own JSON, and the daemon's half by `cargo test`, but CI builds the iOS app without running its tests. Proving the wiring between them is Stage 2's gate.

jiweiyuan:
Rebased onto `main` twice while verifying (it moved 10 commits under me), which changes two of the numbers and inherits one red job that is not mine.

**Green:**

| Check | Result |
| --- | --- |
| `cargo test` | 347 passed, 0 failed |
| `swift build` + `swift test` | 950 passed, 0 failed |
| `cli_compat.py` | **110/110** (was 93/93 — `main`'s clap CLI port added checks; nothing was removed) |
| iOS unit tests | 34 passed, 0 failed |
| CI `linux`, `macOS`, iOS `build` | pass |

The earlier `linux` failure was `pair_state::wss_off_wins_against_a_concurrent_adopter` — "empty pairing token" — which is exactly what #567 fixed. Merging `main` picked that up and `linux` is green.

**Red, and not from this branch: CI `termiod / macos`.**

35 tests fail — `TermiodFilesIntegrationTests` (33) and `TermiodWireOrderIntegrationTests` (2). The same 35 fail on `main`'s own latest `termiod` run (`3f9a9f9`, and `6994755` before it), and the failure sets are byte-identical:

```
comm -13 main-failures mine-failures   →   empty
```

They land in the `Swift client against this daemon` job, which is the only place these integration tests run un-skipped — which is why `swift test` is 950/0 locally. `git diff origin/main` for `termiod/src/files.rs`, `TermiodFiles.swift` and `TermiodFilesIntegrationTests.swift` is empty on this branch; the assertions that fail are about a grep subprocess starting, and #564 replaced `git grep` with ripgrep's crates in-process.

So this is `main`'s breakage from #564/#565, inherited by merging. I have not touched it — it belongs to whoever owns that port, and fixing someone else's regression inside this PR would bury it. Flagging rather than absorbing it. `main` needs a fix before this can merge green; this branch adds no failures to it.

jiweiyuan:
P1 fixed on both halves. `720d3ce`.

You were right that this is the same family, and it turned out to be the same *mechanism* — so the fix is in the shared machinery rather than in `status:` alone.

**Host — `ResourceState::attach`.** Registration and replay now happen in one critical section, with the replay queued onto the subscriber's channel *before* the subscriber is installed. Nothing the method admits can precede it, and a `publish` that wins the lock next necessarily lands behind it.

`SubscribeReply` no longer carries a `replay` field for the caller to send after the lock — that field *was* the bug, so it is gone rather than documented. `fs:` and `git:` rode the same split and are fixed by the same change; the `git:` gap snapshot moved inside the section with everything else. (Lock order is unchanged and still safe: `refresh_git` releases the snapshot before it touches the state, so the two never nest the other way. Noted at the call site.)

Ack-before-replay was already right and stays right, for a reason worth writing down: the ack goes straight to `out` from `process_control`, while the replay reaches `out` through `resource_rx`, which the select loop only polls after `process_control` returns. So the order on the wire is ack → replay → live.

**Client — cursor discipline.** The ack's `seq` no longer touches the durable cursor. Batches arriving before the ack are held; on settle they drain one at a time, and anything arriving mid-drain queues behind them rather than overtaking.

I used `DeviceWatchLedger` rather than inventing an idiom, and **moved it to `Shared/`** to do it. Its own comment says a second copy of the reasoning is how the `fs:` half came to be missing it; the phone is the third plane to need it, and a third copy is how it goes wrong again. The Mac's two planes now use the same file.

One thing your description implied that I had to make explicit — **two marks, not one:**

- `statusCursor` — highest **contiguous** batch applied. This is what goes back as `since`.
- `statusApplied` — highest applied at all, **per attempt**.

They part only across a hole: the batch applies (newer truth beats none) while the cursor stays put so the hole is re-asked for. My first cut used one mark and scenario (b) failed — 11 applied, cursor left at 8, then 9 arrived, tested `9 <= 8`, and overwrote 11. The second mark is what makes the drop correct.

It has to reset to the cursor on every new subscribe, and that is not cosmetic: a fresh subscription replays *in order* from the cursor, so what looked stale under the old attempt is exactly what this one must apply. Without the reset, 9 and 10 are dropped as stale on every reconnect and the cursor is stuck at 8 forever — a slower version of the bug being fixed.

A link drop bumps the generation, so an ack landing on the next connection cannot settle a dead attempt and its held batches cannot apply against a baseline that never installed. `statusCursor` deliberately survives that — it is the point of resuming.

**Regression tests, both scenarios plus the edges:**

Rust (`resource.rs`):
- `a_live_batch_cannot_overtake_the_replay_a_subscriber_is_owed` — publishes live the instant the subscription exists; asserts `[1, 2, 3]` off the subscriber's own channel.
- `the_ack_seq_names_the_end_of_the_replay_it_is_handed`

iOS (`TermiodStatusResourceTests`):
- `testADisconnectRightAfterTheAckDoesNotSkipTheReplay` — scenario (a)
- `testAStaleReplayedBatchDoesNotOverwriteALiveOne` — scenario (b)
- `testABatchArrivingBeforeTheAckIsHeldUntilItLands`
- `testAnAckFromADeadAttemptIsIgnored`
- `testAGapRestartsTheCursorAtTheHostsBaseline`
- `testAnAlreadyAppliedBatchIsDropped`

Three of the round-1 tests started failing once the ledger went in — they had no ack, so their batches were held, exactly as designed. Their fixture now delivers one.

## Verified

- `cargo test` — **349 passed, 0 failed**
- `swift build` clean, `swift test` — **950 passed, 0 failed**
- `cli_compat.py` — **110/110**
- iOS unit tests — **38 passed, 0 failed**; simulator build clean

One environment note, because it cost me a false signal in round 1 and nearly a second one here: this worktree had `termiod/target` symlinked to the main checkout, and a build in another checkout kept clobbering shared artifacts — first a stale `termio` binary that read as a `cli_compat` regression, then a stale `termiod-vt` rlib that reported `screen_text` missing when it is on line 569. It has its own `target/` now and builds from scratch. Worth knowing if anyone else works this repo from a worktree.

CI `termiod / macos` is still red for `main`'s 35 pre-existing failures (#564/#565), unchanged and unrelated.

jiweiyuan:
P1 fixed, note addressed, and main's #572 merged — which is what was actually killing the macos job. `992aa37`.

## P1 — Mac `git:` cursor

Confirmed at both ends of the path: the ack set `gitCursor = max(cursor, seq)` before any replay applied, and `apply` used `max` rather than contiguity, so a hole advanced it too.

Rather than write the rule a third time, it is now a shared value type — `ResourceCursor` in `Shared/`, alongside `DeviceWatchLedger`. The two are complementary and neither implies the other: the ledger orders a batch's *arrival* against the ack, the cursor decides what that batch is *worth* once it has arrived. Three planes, one rule, nine tests (`ResourceCursorTests`) covering both your scenarios plus the deadlock the per-attempt reset prevents.

`git:`'s gap case needed one thing the phone's did not: the device synthesizes a **full** state at its current seq for a gap subscriber, so that batch is adopted outright rather than walked to. Without it, the cursor would never reach the snapshot's seq and every reconnect would re-snapshot forever.

**`fs:` — checked, and it does not have the bug.** It never takes the cursor from the ack: the ack's seq goes to the caller as `.established(at:)`, and `self.cursor` moves only in `deliver`, for batches that applied — the comment there already says why. What it did share is the weaker property, `cursor = batch.seq` jumping a hole, so it now follows the same contiguity rule in place. I left it inline rather than converting it to the type: it is freshly merged from #561, it is under its own lock, and `noteListingCursor` legitimately adopts a listing's stamp as a baseline. Converting it is a follow-up, not something to do inside this PR.

iOS `status:` moved onto the shared type, dropping its own copy of the same three lines. All 38 iOS unit tests still pass unchanged, which is the check that the extraction was faithful.

## Note — the gap replay under the lock

Looked at moving it out. **It cannot move, and the reason is a correctness one rather than effort:** the full batch must describe the snapshot *at* the seq it is stamped with. `refresh_git` writes the snapshot and then publishes, so a build reading the snapshot outside this lock could be stamped with a seq taken after that publish — old state under a new cursor, and the delta it missed skipped for good. Holding the state lock across both reads is exactly what makes that interleaving impossible.

So it is the comment, with the bound stated: `cap_statuses` caps a snapshot at `STATUS_CAP` (5,000) entries and `STATUS_PATH_BYTES_CAP` (1 MiB) of paths before it ever reaches here, so the clone-and-sort is a five-thousand-element sort at worst, on a lock no other resource shares. A tree big enough to matter is already cut.

## The macos job

You were right to push on this, and my round-2 reasoning about it was wrong. I had compared "failing test names" between my run and main's using a grep that matched **every test name in the log**, not just failures — so the sets were trivially identical and the comparison proved nothing.

Running the job's actual command locally (`TERMIO_TERMIOD_TEST_BIN=… swift test --filter …`, which is why a plain `swift test` skips 61) showed the truth: **7 failures in 2 tests**, not 35, both about the `git grep` child that #564 replaced with in-process ripgrep. Main fixed exactly that in #572 with a stall seam. Merged, and the job is green locally:

```
Executed 35 tests, with 0 failures
```

## Verified

| Check | Result |
| --- | --- |
| `cargo test` | 349 passed, 0 failed |
| `swift test` | 959 passed, 0 failed |
| Real-daemon job, locally | **35 passed, 0 failed** |
| `cli_compat.py` | 110/110 |
| iOS unit tests | 38 passed, 0 failed |
| iOS simulator build | clean |

Branch is level with `main`. Waiting on CI to confirm.

PR #575, opened 2026-09-01
Decouple PTY size from the write token

jiweiyuan:
Implements §4 of [PTY size is not the write token](docs/design/20260901-pty-size-is-not-the-write-token.md): the PTY's size stops riding on the write token and becomes a session-level policy the daemon computes over the attachments actually rendering the session. The token itself is untouched — it still gates input, still moves by typing, and observers are still refused.

## Protocol (both ends of the wire)

- `R` is now a **viewport declaration** accepted from any attachment: rows, cols, plus an optional fifth byte for a surface that is hidden without detaching. The four-byte form still means "rendering", so the common case is byte-identical to v0.
- The fifth byte is behind a negotiated `viewport` capability. A daemon that predates it rejects a five-byte resize as malformed and drops the attachment, and the daemon grants only what a client asks for — so a client that does not see it granted keeps to the four-byte form and simply has no hidden-pane filter, exactly as before.
- `Control::Attach` carries whether the attaching surface is on screen. Assuming `true` put a pane that mounted hidden into the min for one message, which cost every other viewer a PTY resize and a barrier repaint before its own next frame undid it.
- The daemon sizes the PTY to the **per-axis min** (rows and cols independently, Zellij 0.45 / zellij#5133 style) over the attachments with `rendering = true`. Observers are excluded: they have no surface, and `termiod observe` attaches at a placeholder 24x80. An empty rendering set keeps the last size. When the min moves, the existing barrier runs unchanged.
- The `resize_claim` control frame is deleted; promotion is announced by `writer_changed` alone. A frame from an older daemon decodes as unknown and is ignored.

## Client side

- `applyWriter`'s grid re-assert on token grant — deleted, Mac and iOS. This was the line that turned one misclassified byte into a 6135-moves-in-30s resize storm.
- `applyAuthoritativeGrid`'s writer-answers-divergence branch — deleted, both ends.
- The iOS scale-down letterbox — deleted. Under min an observer is never *narrower* than the PTY, so there is nothing to scale.
- The Mac's `SharedGridLetterbox` is **kept**, with the half that actually had to change. Its job was never only scaling: it lays the surface out at the shared grid, without which the surface parses bytes wrapped for a narrower PTY at its own width — the §C.5 divergence this RFC's own §4 promises to avoid ("both surfaces sit at the shared grid"). What could not survive was feeding that back: a letterboxed surface must not declare the grid it was letterboxed to, or this pane's viewport becomes the min and pins the session at the narrowest size any viewer ever had. Layout grid and declared viewport are now separate — the surface sits at the shared grid, the pane declares its own capacity — and the letterbox no longer gates on the write token, because under min any attachment can be wider than the PTY.
- `TermiodClient.resize()` / iOS `resize()` lose their writer gates; a client-side dedupe keeps redundant frames off the wire.

## Visibility, both devices

The Mac plumbs the pane-visibility bit it already computes for `setSurfaceVisible` into the link, so fifteen background panes cannot pin the grid. The phone parks a screen rather than tearing it down — socket, stream and surface all stay alive — so it sends a `rendering` companion message on park and on restore. Without that half, a phone that merely swiped back to the session list held the Mac at phone width for as long as the app stayed up.

## Tests

Daemon: per-axis min across attach / detach / hide / show; empty-rendering-set retention; reattach at a different size delivers the barrier keyframe (the zellij#5373 seam); an observer never moves the size and does not strand the session on its way out; a hidden attachment stays out of the min until it is shown.

The smoke test drives a real daemon over the wire and had two checks still asserting the old design — "the winsize follows the token" and "promoted writer receives a resize claim". Both now assert the new invariant. The unit suites could not have caught these.

`alternating_typists_move_the_token_but_never_the_size` passes on `main` too: the storm ran through a *client*, which a daemon-only test cannot observe. Its docstring now says what it actually covers rather than claiming the storm.

Also fixed: a resize the PTY refused was reported `retryable` but could not be retried, because the declaration was stored before the policy ran and the retry deduped against itself.

## Notes

The `ESC[I` classifier fix landed separately on `main` in #576; `origin/main` is merged in here.

Not yet exercised on hardware: the two-attachment path — a phone and a Mac on one session, the letterbox rendering at the shared grid, and the phone releasing the Mac on park — needs a paired device, per §4 of the bug report.

## Release Notes:

- Opening a session on the phone no longer fights the Mac over the terminal's size: the grid is sized to fit every device actually viewing the session, and typing on either end moves only the keyboard, never the layout. A pane hidden behind another, or a session you have swiped away from on the phone, no longer constrains the grid at all.


jiweiyuan:
Superseded. The size policy landed as #580, #586 and #589; the backgrounded-phone half as #590. The repaint flag is off the write token in main too, as `repaintPending` rather than `surfaceBehindSharedGrid` — same semantics, different name, which is why it did not show as already-applied.

PR #578, opened 2026-09-01
Derive PTY size from attachments

jiweiyuan:
Derive the PTY grid from the minimum declared interactive attachment viewport, independently of the write token.

Caveat: before the client clean-up lands, a token-holding Mac pane that is not the smallest viewport can briefly render its shared grid without letterboxing and may discard a barrier keyframe; live bytes repair it.

Release Notes:
- Keep shared terminal sessions at a stable size while writers change.

### DIFF
--- a/termiod/src/client.rs
+++ b/termiod/src/client.rs
@@ -3,7 +3,7 @@
 
 use crate::paths;
 use crate::protocol::{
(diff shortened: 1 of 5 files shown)


jiweiyuan:
Superseded by #580, which merged this same commit (26d44fa) plus the smoke-test rewrite CI needed — `the winsize follows the token` asserted the invariant this change removes.

Issue #609, opened 2026-09-05
Reopening the app repaints agent statuslines as a sparse, scattered frame; daemon VT model is correct


kirbyzhu:
### What happened

Every time I quit termio (`Cmd+Q`) and reopen `termio.app`, the **multi-line statusline of every running agent pane repaints as a sparse, scattered frame** on the first render after reattach. It affects two independent programs simultaneously — Claude Code (a custom multi-line statusline) and Codex (a ~15-line `codex-hud` status_line panel). Almost the whole statusline area is left blank; only a handful of narrow (ASCII) glyphs survive, and they land at the wrong coordinates. The garbled frame **persists** until something forces a full repaint (typing, scrolling, or resizing the pane) — it does not self-heal on its own.

In my words: termio 每次重新打开后，Claude Code 和 Codex 面板的 statusline 都是乱序。

### Evidence that isolates the layer (daemon VT model is correct; the app's paint is wrong)

At the exact moment the screen was garbled, I read the daemon's authoritative VT model for both panes:

```
termio sessions read <codex-id>
termio sessions read <claude-id>
```

**Both `read` outputs were complete and correctly ordered** — the full Codex HUD panel and the full Claude Code statusline, every CJK label and every value in the right place. So termiod's snapshot (the `S` frame machinery from #181) is faithful.

The **screen**, at the same instant, showed the opposite: for the Claude pane, only the top border row rendered; below it the grid was essentially empty except for scattered digits (`9  94`, `7`, `8`, `8`, `2`, `2`, `6`, `5 42 4 5`) — these are the numeric values (9% / 88% / 25% / 33 GB / counts …) left over from statusline rows whose CJK-heavy label cells were never painted. For the Codex pane, the entire multi-line HUD panel was replaced by blank grey blocks.

So: **snapshot data correct → on-screen paint incomplete and mis-placed.** The bug is in the app applying the snapshot to the ghostty surface on cold reattach, not in the daemon.

The signature (mostly-blank grid, a few narrow glyphs at wrong columns, wide/CJK cells dropped) points at a **cell geometry / column-stride / wide-char miscount** in the reattach repaint path — the same family as #306 ("painting rows from column 73") and the `rows/cols` authority that #181 added to the `S` frame.

### Likely contributing factor: app/daemon version skew

`termiod` is durable and survives app quit, so an app upgrade does **not** restart it. On this machine:

```
termio.app     0.49.0+1881   (running)
[...]

PR #618, opened 2026-09-07
Bring this Mac's termiod current when the app updates

jiweiyuan:
Implements `docs/design/20260906-app-install-updates-termiod.md`: updating the app staged a new `termiod` and left the running one alone forever, so the machine the app runs on was the only one it talks to that never got current (#609 sat two releases behind). The local launch reconcile now takes the same deploy ladder every remote already gets — automatic while it is free, asking only at the rung that costs a session.

## What changed

- **`roll_back()` hands off before it stops** (`termiod/src/lifecycle.rs`). An unhealthy new image is handed back to `termiod.prev` — same pid, every PTY carried — and the destructive `stop --force` + `mv` survives only as the fallback when the handback itself fails. The previous build's own CLI drives the recovery, since the staged binary just failed verification. This lands first so the failure path is non-destructive before anything fires automatically.
- **`reconcileLocalDaemon()` drops `--stage-only`** (`Sources/termio/Terminal/Termiod/TermioStore+Termiod.swift`). At launch, a stale local daemon is handed the bundled binary through `deploy --json`: `current` (the daemon handed off in place) logs and asks nobody, because nothing was at risk; `staged` (a daemon too old to hand off, or a handoff that failed) raises the same names-what-would-be-interrupted dialog the machines pane uses, where Update Anyway is a deliberate destructive click and Not Now owns Return and Escape; `unhealthy` logs what rollback did. The staged-message construction is shared with `performRemoteReadyCheck` rather than written twice. The `bundledStamp == desired` guard is untouched — a dev build still never deploys over a released daemon.
- **`termio version` names the local skew** (`termiod/src/version.rs`). A running local daemon older than the binary staged at its path was the one skew in the table with no command beside it; the row now reads `← behind; run `termiod handoff``.

## Verified

- `cargo test` in `termiod/` — all suites green, including the new `an_unhealthy_new_image_is_rolled_back_without_stopping` (asserts a `handoff --binary` and no ` stop` anywhere) and the reshaped `a_daemon_that_does_not_come_up_is_rolled_back` (handback fails → `stop --force` fallback).
- `swift build` and `swift test` — 996 tests, 0 failures.
- The version hint, against the real skew on this Mac (app 0.50.0+1913 over daemon 0.49.0+1881): the row prints `termiod local  0.49.0+1881  proto 1  ← behind; run `termiod handoff``.
- Not yet exercised end to end: the launch reconcile against a live older daemon needs the next release install (a dev build refuses to deploy over a released daemon by design). The design doc's §9 manual check applies to the first tagged build carrying this.

Release Notes:
- Updating Termio now updates this Mac's session host too: the daemon takes the new binary in place, keeping its pid and every running session. When it can't do that for free, Termio names the sessions in the way and asks instead of interrupting them.
- `termio version` flags a local daemon running behind the installed binary and says how to bring it current.


jiweiyuan:
Review round 1: 4 findings confirmed and fixed in 8f422c03..063ef549 — launch deploys are handoff-only (an idle session is never stopped unasked), deploy subprocesses are pinned to the app's own channel/socket, the local node keeps a durable rollback stash so recovery actually engages, and the version hint names the resolved daemon path.

jiweiyuan:
Review round 2: 4 more findings, fixed in 4449030b + 3c0c00cc — the empty-daemon bounce re-reads the roster after the failed handoff (and a pre-handoff daemon can't be asked atomically, so the residual read-to-stop gap is documented), a stash-based rollback reports success only when a daemon actually took the handback (never stops, since autostart would revive the failed build), the stash is written via copy-then-rename, and the version hint carries TERMIOD_SOCK/TERMIO_CHANNEL.

Issue #638, opened 2026-09-13
Multi-second main-thread stall on workspace switch; mounted panes and surfaces accumulate with no eviction


jiweiyuan:
## What happened

Heavy beachballing through a normal working day on 0.53.2, worst when switching workspaces or sweeping through sessions. A user report describes the same curve from the outside: fast right after install, sluggish after a few days of use.

## Diagnosis

**1. Workspace switches stall the main thread for seconds, and the cost scales with accumulated sessions.** The trace log:

```
17:05:23.215  workspace switch to=Work  elapsed_ms=0.83
17:05:26.587  workspace column          elapsed_ms=3371.95
```

The `workspace column` span measures how long a queued main-thread callback waited to run (`TermioStore+Workspaces.swift:374`) — 3.37 s of the run loop spent applying `currentWorkspaceID`: SwiftUI reconciling the hosted sidebar and pane tree, then the display cycle committing. A 150 ms stall of the same shape hit two minutes earlier. The stall's own stack is uncaptured — the `sample` ran three seconds after it ended and shows a healthy idle main thread — so the split between SwiftUI reconciliation and anything else inside that turn is not yet proven.

**2. What is proven is unbounded retention, which explains both the stall size and "fast at install, slow after days".**

- No capacity, age, or LRU eviction exists for still-open sessions. Cached surface state clears only on project close (`TermioStore+ProjectActions.swift:1215`), session close (`:1366`), or relaunch (`:1436`).
- Once activated, a terminal pane stays mounted across workspaces: `activated` only accumulates IDs, `mounted` filters on session existence, not workspace (`TerminalPane.swift:74`, `:361`); hidden panes are drawn at opacity zero (`:195`).
- This state file is a few days old and carries 43 sessions across 5 workspaces. A `sample` of the app shows 16 retained surfaces — 16 renderer + 16 io + 16 cf_release threads, 103 threads total, 918 MB — for a window showing a handful of panes. Hidden surfaces do park their display link (`TerminalSurfaceCoordinator.swift:288`), so the cost is memory plus per-update view reconciliation over an ever-growing mounted set, not idle rendering.
[...]

PR #640, opened 2026-09-13
Capture main-thread stalls in the act with a sampling watchdog

jiweiyuan:
Step 0 of #638. The 3.4 s workspace-switch stall was diagnosed entirely from timing spans; by the time anything could sample the process, the main thread had recovered. This adds the missing instrument: a utility-queue heartbeat posts an acknowledgement to the main queue, and when it goes unanswered for 500 ms a child `/usr/bin/sample` records the process while the stall is still happening.

- One capture per stall (re-armed only by the acknowledgement finally running), a 60 s cooldown, and five rotating files under `Application Support/termio/stall-samples/`.
- Sleep/wake drops the pending acknowledgement so a parked main thread never reads as a stall.
- `sample` failures are written into the capture file — a signed release refusing task inspection should say so, not look like the watchdog never fired.
- Off by default: `defaults write sh.termio.app StallCaptureEnabled -bool YES`.
- The `workspace column` span now reports the retained-surface count, so the line that says how long the turn took also says how much mounted state it reconciled.

`swift build` and `swift test` pass.

Release Notes:
- Nothing user-facing. Adds an opt-in diagnostic that samples the app during main-thread stalls, for #638.


jiweiyuan:
Parking this — it's diagnostics-only and can't reproduce the stall synthetically (dev-channel sessions reconcile in ~2 ms; the real stall needs the heavy agent workload of a daily-use release). Not worth carrying in a release just to enable a manual debug flag. Branch fix/stall-watchdog stays; reopen when the retention-eviction work needs the captured stack. Context: #638.

PR #663, opened 2026-09-18
Fix the file editor drawing unhighlighted text in black on dark backgrounds

### LINKED ISSUE #662: File editor draws unhighlighted text in black on dark backgrounds
### What happened

In a dark appearance the file editor draws text that gets no syntax highlighting in **black**, which is unreadable against termio's dark terminal background (`#212121`) — a **1.14:1** contrast ratio.

`LICENSE` is the easiest place to see it: it has no extension, so `FileEditorView.highlightLanguage(for:)` returns `nil` and the buffer never receives a highlight pass. The same happens to any file past the 256 KB highlight limit (`highlightDisabled`), and to any language highlight.js has no grammar for.

### Steps to reproduce

1. Set Appearance to **Dark** (or run the OS in dark mode).
2. Open a project whose repository has a `LICENSE` file.
3. Click `LICENSE` in the file tree.
4. The body text is black on the near-black editor background. The line numbers stay legible — they draw through `AppSettings.gutterInk`, which does resolve against the theme.

The file reads as empty at a glance; the text is only visible where the screenshot is brightened.

### Root cause

`HighlightedTextView` seeds every glyph with `baseAttributes`, and that dictionary carries only the *layout* metrics — font, paragraph style, baseline offset. It carries no `.foregroundColor`, and nothing else sets one on the seeded text.

A highlight pass is the only thing that ever colors the text: Highlightr's parser strips `.hljs`'s own background but keeps its `color` declaration, and its per-token rules bring their own color. A file with `language == nil` never enters that path at all, so the storage keeps whatever AppKit picks for an `NSTextView` whose text carries no color — **black, regardless of appearance**. `NSTextView().textColor` is `#000000` under both `aqua` and `darkAqua`, and it also cannot simply be assigned on every update, because `NSText.color =` is a whole-document setter that would wipe the highlight colors.

Because the editor paints an opaque terminal background rather than `textBackgroundColor`, black does not fail loudly in light mode either — it just happens to be correct there.

With `xcode-dark`, a file that *does* highlight is unaffected (`language: "swift"` renders correctly), which is what makes this look like a font or theme bug rather than a missing base color.

Related: the caret falls back to `NSColor.textColor` (`FileEditorView:183`) when no Ghostty theme is selected. That one is *not* broken — `NSTextInsertionIndicator.color` does resolve a dynamic system color per appearance (verified: white under `darkAqua`, black under `aqua`), it just takes the system ink rather than the theme's accent. Left as-is.

### Evidence

`LICENSE` open in the editor, dark appearance (termio 0.53.3, macOS 26.6.2). The line numbers are legible; the body text is not:

LICENSE rendered with black ink on the dark editor background

Measured on that screenshot of `LICENSE` open in the editor (dark appearance, termio 0.53.3, macOS 26.6.2), body region only, background `#212121`:

| | count |
| --- | --- |
| pixels **darker** than the background | 28,312 |
| pixels **brighter** than the background | 0 |
| brightest pixel in the text column | `#212121` (the background itself) |
| brightest pixel in the gutter | `#9B9B9B` (the line numbers) |

Zero pixels above the background means the glyphs are literally painted black: the line numbers are the only readable ink on screen.

Verified in isolation against the same `HighlightedTextView` setup (theme `xcode-dark`, background `#212121`), measuring the brightest pixel produced:

| buffer | brightest |
| --- | --- |
| `language: nil`, no base color (today) | `#2C2C2C` — invisible |
| `language: nil`, base color = theme foreground | `#FFFFFF` |
| `language: "plaintext"` | `#FFFFFF` |
| `language: "swift"` | `#FFFFFF` |

### Affected code

- `Sources/termio/Editor/HighlightedTextView.swift:306` — `baseAttributes` (no `.foregroundColor`)
- `Sources/termio/Editor/HighlightedTextView.swift:314` — `apply(to:)` (`font` is change-gated; `backgroundColor` and `insertionPointColor` are assigned unconditionally)
- `Sources/termio/Editor/FileEditorView.swift:336` — `language: highlightDisabled ? nil : language`
- `Sources/termio/Editor/FileEditorView.swift:183` — the same appearance-blind `.textColor` fallback for the caret

### Expected

Plain, unhighlighted text is drawn in the ink the rest of the editor's chrome uses — the selected terminal theme's foreground, or the system label color when no theme is picked — matching how `backgroundColor` and `lineNumberColor` are already resolved in `FileEditorView`.

### Notes on a fix

I have a patch ready on a branch, which:

- adds an `AppSettings.editorInk(for:)` next to `gutterInk(for:)` — the theme's own foreground, falling back to `ChromeTheme.overlayInk` at full alpha when no theme is picked;
- threads that through `HighlightedTextView` and into `baseAttributes`;
- keeps the existing fallback honest: the *system* `NSColor.textColor` is not a substitute here, because it tracks the OS appearance rather than the terminal theme (a warm cream or a tinted grey foreground would be replaced by plain black/white).

One thing worth flagging, because it looks like a bug and is not: **22 of the 485 bundled themes pair a foreground below WCAG AA with their own background** (9 below 3.0 — Darkermatrix is 1.89:1, C64 2.26:1). Using the theme's foreground reproduces exactly what that theme renders in the terminal, which is the right behavior; a contrast floor in the editor would make it disagree with the pane behind it. My regression test asserts identity with the theme foreground, not a contrast floor — only the *no-theme* fallback is held to AA, on the canvas it is actually drawn on.

Happy to open a PR against `termio-sh/termio` if you want it.

jorben:
## What

Fixes #662 — in dark mode the editor drew a plain-text file's body in black on the `#212121` canvas, a 1.14:1 contrast ratio, so a `LICENSE` read as an empty file.

`highlightLanguage(for:)` returns `nil` for an extension-less file (and `highlightDisabled` forces the same for anything past the 256 KB limit), so the buffer never gets a highlight pass. Only that pass ever colored this text — the seeded attributes carried layout metrics alone, leaving AppKit's built-in black to draw the glyphs.

The base ink is now part of the seed, resolved the way the gutter already resolves its numbers:

- `AppSettings.editorInk(for:)` — the selected theme's own foreground, falling back to `ChromeTheme.overlayInk` at full alpha when no theme is picked.
- `HighlightedTextView` takes it as `textColor` and puts it in `baseAttributes`, so it is on every glyph from the first frame and survives buffers the highlighter never touches.

Two things worth calling out:

- The `textView.textColor` write is **change-gated** like the existing font one. `NSText.color` is a whole-document setter, so assigning it unconditionally on each update pass would re-stamp plain ink over every syntax color on every keystroke.
- `NSColor.textColor` is deliberately *not* the fallback. It tracks the system appearance rather than the terminal theme, so a warm or tinted theme foreground (Catppuccin, Gruvbox, Rose Pine) would be flattened to plain black or white and stop matching the pane behind the editor.

Nothing changes for files that do highlight: the pass still overwrites the color.

## How it was verified

**Not fully — `swift test` did not run locally, and the PR needs CI to confirm.** This machine has Xcode 26.2 while `.github/workflows/macos.yml` pins `DEVELOPER_DIR` to 26.4, and 26.2 emits 20 `SendingRisksDataRace` diagnostics in `CompanionServer.swift` that 26.4 does not. They are present on unmodified `main` and block the whole target from compiling here; CI is green on `main`. No diagnostic in the build log points at any file this PR touches.

What I could check:

- `swiftlint lint` — 0 violations across all four files.
- The changed sources compile: the build log's only errors are the 20 `CompanionServer.swift` ones, none in `Editor/` or `Theme/`.
- `Tests/termioTests/EditorInkTests.swift` parses cleanly (`swiftc -parse`).

**Pixel probe** against the real failure, driving `CodeAttributedString` + the `xcode-dark` theme on the `#212121` canvas exactly as `HighlightedTextView.makeNSView` does, measuring the brightest pixel produced:

| buffer | before | after |
| --- | --- | --- |
| `language: nil` (a `LICENSE`) | `#2C2C2C` — invisible on `#212121` | `#FFFFFF` |
| `language: "swift"` | `#FFFFFF` | `#FFFFFF` — colored path unchanged |

**Mutation check** on the new assertions' logic: swapping the broken implementation back in takes them from 4/4 to 2/4 (`darkFallbackIsLight` and `darkClearsAA` fail), so they do catch this regression rather than passing vacuously.

I also checked the fix against all 485 bundled themes before writing the tests, which turned up something worth a look: **22 themes pair a foreground below WCAG AA with their own background** (9 below 3.0 — Darkermatrix 1.89:1, C64 2.26:1, Royal 2.34:1). Using the theme's foreground reproduces exactly what that theme renders in the terminal, which I believe is correct — a contrast floor in the editor would make it disagree with the pane behind it. So the test asserts *identity* with the theme foreground, and only holds the no-theme fallback to AA. Flagging in case you'd rather clamp; it is a one-line change and a product decision, not a bug.

## Release Notes

- Fixed the file editor drawing unhighlighted text in black on dark backgrounds, which made extension-less files such as `LICENSE` look empty.


jiweiyuan (dismissed):
The base ink is right, but the gate in `apply(to:)` doesn't hold: `NSTextView.textColor`'s getter returns the **first character's** foreground color, not the ink last assigned — so on any highlighted file it never matches, and the whole-document setter fires on every `updateNSView` pass.

Against the real `CodeAttributedString` + `Highlightr` (`import Foundation\nlet x = 1`, `xcode-dark`):

```
distinct colors before:     3
textView.textColor getter:  sRGB 0.988 0.373 0.639   ← the `import` keyword
distinct colors after:      1
```

`processEditing` only re-highlights on `.editedCharacters`, so an attribute-only write triggers nothing and the colors stay flattened.

The line isn't needed. An appearance switch flips `appliedTheme`, and that path re-applies `baseAttributes` — carrying the new ink — across the whole document before re-highlighting. Only newly typed characters are left over, and those want the typing attributes:

```diff
-        if textView.textColor != textColor { textView.textColor = textColor }
+        textView.typingAttributes[.foregroundColor] = textColor
```

With that change `swift build` is clean and all 1029 tests pass here. Worth one test over this gate as well — it's the case the new suite doesn't reach.

The rest holds up. I rendered the seeding as it stands today on the `#212121` canvas and the brightest pixel in the view is `#212121`, so the report is if anything understated. Your call on the 22 sub-AA themes is the right one: matching the theme is the point.

jiweiyuan:
Merging this — the base ink is right and the bug is worse than the report says (I rendered the current seeding on `#212121` and the brightest pixel in the view *is* `#212121`). The `textColor` gate goes in a follow-up: your fork is org-owned, so GitHub won't let a maintainer push to the branch. Thanks for the write-up, the diagnosis was exact.

PR #666, opened 2026-09-18
fix: paste absolute paths for files copied from Finder

### LINKED ISSUE #625: Pasting a file should insert its absolute path instead of only the filename
Termio has become my primary terminal environment for agentic development.

Its project and worktree management make it practical to keep multiple tasks isolated and organized, and the Pi integration is especially useful in my daily workflow. The product already handles the larger agent-development workflow very well, so this small input behavior stands out when working with local files.

### Problem

When I copy a local file in Finder and paste it into a Termio chat or agent input field, only the filename is inserted.

For example, copying this file:

```text
/Users/<username>/Desktop/example image.png
```

currently inserts:

```text
example image.png
```

This loses the file location, so the agent or terminal cannot reliably identify or access the intended file.

### Steps to reproduce

1. Copy a local file from Finder.
2. Focus a Termio chat or agent input field.
3. Press `Cmd + V`.
4. Observe that only the filename is inserted.

### Expected behavior

Pasting a local file should insert its absolute path, for example:

```text
/Users/<username>/Desktop/example image.png
```

For paths containing spaces or shell-special characters, the inserted value should be quoted or escaped so it can be used directly by an agent or in a shell command.

### Suggested direction

When the clipboard contains a file URL, preserve its standardized absolute path instead of reducing it to the last path component (the filename).

### Related issue

This is related to #610, but it is a different case:

- #610 concerns dragging files from Finder into the chat input.
- This issue concerns pasting a copied file and receiving only its filename.

### 问题描述

Termio 已经成为我主力的 Agent 开发终端。

它的项目管理和 Worktree 管理功能让我可以方便地隔离、组织多个任务；同时，对 Pi 的支持也非常好，已经融入我的日常开发工作流。正因为整体的 Agent 开发体验已经很成熟，处理本地文件时这个输入细节会显得比较突出。

目前在 Finder 中复制本地文件后，粘贴到 Termio 的对话框或 Agent 输入框里，只会插入文件名，而不会插入文件的完整路径。

例如复制：

```text
/Users/<username>/Desktop/example image.png
```

当前粘贴结果为：

```text
example image.png
```

文件位置信息因此丢失，Agent 或终端无法可靠地定位和访问目标文件。

### 复现步骤

1. 在 Finder 中复制一个本地文件。
2. 聚焦 Termio 的对话框或 Agent 输入框。
3. 按下 `Cmd + V`。
4. 观察到输入框中只插入了文件名。

### 期望行为

粘贴本地文件时，应插入该文件的绝对路径，例如：

```text
/Users/<username>/Desktop/example image.png
```

如果路径中包含空格或 Shell 特殊字符，建议自动进行合适的引号包裹或转义，使其能够直接被 Agent 或终端命令使用。

### 建议方向

当剪贴板内容是文件 URL 时，建议保留其标准化后的绝对路径，而不是仅取最后一个路径组件，也就是文件名。

### 相关 Issue

此问题与 #610 有关，但不是同一个问题：

- #610 是将 Finder 文件拖入对话框后，无法正确识别文件路径。
- 本 Issue 是复制文件后粘贴时，只保留文件名、丢失完整路径。

mvanhorn:
## What

Extend the existing paste interception in `TermiodTransfer.swift` to recognize actual file URLs before the current remote-image branch or ordinary text fallback, using a small pasteboard reader in that file which the production handler calls. Read file-only NSURL objects with AppKit, standardize each file URL, convert it with the existing `TermioStore.promptToken(for:)`, and join the quoted tokens with spaces; never infer file identity from clipboard text or use `lastPathComponent`. Send the resulting text to the resolved session surface using the same `surface.send` path used by `TerminalPane.sendPaths`, without modifying the system clipboard, synthesizing Return, or adding unconditional bracketed-paste escape sequences.

Copying a file in Finder and pressing Command-V in a Termio agent prompt inserts only its filename, which loses the location the agent needs. The reporter confirmed that the affected agent is Pi after a collaborator asked for that detail. File pastes should insert standardized absolute paths with shell quoting for spaces and special characters. This issue concerns clipboard paste; changes to Finder drag-and-drop in #610 are outside its scope.

Fixes #625

## How it was verified

On an isolated NSPasteboard, provide both a file URL and filename text for `/Users/example/Desktop/example image.png`; assert that the production reader returns the shell-quoted absolute path, not the basename or a percent-encoded URL.
Cover multiple file items in order, directories, Unicode, spaces, apostrophes, dollar signs, backticks, semicolons, and lexical `..` normalization. Assert that each path remains one quoted argument and no Return is appended.
Cover a file URL accompanied by TIFF/PNG/icon data: choose its path and do not upload the image representation. Preserve existing image-only and text-plus-image regression tests.
Empty pasteboards, ordinary text (including text that looks like a path), non-file URLs, and malformed file data should produce no file-path payload and retain the ordinary fallback without crashing.
Run `swift build` and `swift test --filter ClipboardTransferTests`, then the full `swift test` suite. These are downstream implementation checks; this planning pass does not build or mutate the checkout.
Rebuild using the `macos-rebuild-dev` skill and inspect the running app using `app-screenshot-debug`. Copy actual files from Finder and verify Command-V, terminal context-menu Paste, and Edit-menu Paste in Pi and a shell. Check the delivered absolute paths, one insertion per action, and no automatic submission.
Verify paste into a clicked inactive split reaches that split, non-terminal text fields retain ordinary paste and menu enablement, and modified Command-V shortcuts remain untouched. Check screenshot paste still follows existing local and remote behavior; failed or unattached terminal delivery must not silently report success.

## Release Notes

Nothing beyond what is described above.

AI was used for assistance.


mvanhorn:
Thanks for the merge, @jiweiyuan. Putting the file-URL branch ahead of the remote-image one keeps the existing paste behavior untouched for everything else.

PR #669, opened 2026-09-19
Preview Excalidraw drawings in the editor

jiweiyuan:
Open a `.excalidraw` file and you get the drawing, not the scene JSON.

The Edit/Preview toggle Markdown already has now covers drawings too, defaulting to the picture. All four containers Excalidraw writes decode to the same scene — `.excalidraw`, `.excalidraw.json`, and the doubled `.excalidraw.svg` / `.excalidraw.png`, which previously opened in Quick Look as flat art and now render live at the app's theme. A PNG-embedded scene has no source face, so it previews without the toggle.

## How it works

`ExcalidrawRenderer` is `MermaidRenderer`'s shape, for the same reason: Excalidraw lays a drawing out against a DOM, so it can't run in a `JSContext`. It gets one offscreen `WKWebView` — in a borderless window that is never ordered in (#348) — and hands back finished SVG, checked by the same inertness gate and cached by content hash and appearance.

The bundled engine is the export path only: `exportToSvg` plus `loadFromBlob`, no editor and no save-back. Three decisions are worth calling out, each found by measurement rather than reasoning:

- **Fonts are declared by the page, never inlined per drawing.** `exportToSvg` otherwise awaits a font fetch that never resolves offline — the render simply hangs. The subset faces ship as resources and are embedded once as `@font-face`, the way the Markdown reader already ships iA Writer.
- **The container is chosen by sniffing the bytes, not the extension.** Offering a 145KB PNG to the JSON and SVG decoders takes ~40s to be rejected. A test pins the deadline.
- **No background rect.** In dark mode Excalidraw applies `invert(93%) hue-rotate(180deg)` to the whole SVG, so a canvas painted here came back as a light slab on a dark page. The preview page is the canvas instead, which is also what makes it sit flush with the terminal.

## Size

The published package entry statically reaches the editor's 57 UI locales and the mermaid-to-excalidraw importer (mermaid, cytoscape, katex), none of which the export path calls. Stubbing both takes the bundle from 8.2MB to 2.9MB — under the 3.5MB mermaid already costs. Fonts add 350KB; the 12MB Xiaolai CJK face is left out, since macOS resolves CJK through a system fallback. `Excalidraw-README.md` records how to rebuild it.

## Verification

- `swift test` — 1056 tests, 0 failures (11 new).
- The renderer is driven end-to-end from a test through the real WebKit harness: a scene, a scene embedded in a PNG, and a file holding no scene. Verified against a packaged `.app` that every resource resolves through `Bundle.termioResources`.
- Light and dark were checked by rendering the real `ExcalidrawReaderRenderer` output and reading it back.

Not verified: the SwiftUI wiring itself (toggle visibility, mount/unmount) was not seen on screen — Screen Recording is denied to this host, so the app window could not be captured.

Release Notes: Excalidraw drawings now preview in the editor. Opening a `.excalidraw` file shows the drawing instead of its scene JSON, and `.excalidraw.svg` / `.excalidraw.png` render as live drawings at the app's theme.


jiweiyuan:
Not shipping this — termio stays small, and an embedded drawing editor isn't a necessary entity. Branch `feat/excalidraw-preview` is kept if it's ever wanted.

PR #682, opened 2026-09-20
Stop a snapshot painting a background the screen never held

jiweiyuan:
Re-attaching, resyncing or resizing a session restored the screen through libghostty's `Vt` formatter, which emitted a run of blank cells under whatever style the preceding run left active rather than under the blanks' own.

Claude Code's banner is the case that exposed it. It sets a literal black background for its block art, jumps the cursor past three untouched columns with CHA, and only then restores the default:

```
ESC[38;2;215;119;87m ▐  ESC[48;2;0;0;0m ▛███▛█  ESC[12G  ESC[39m ESC[49m ESC[1m Claude Code
```

Nothing paints columns 9–11 on a live stream. The formatter re-serialised the gap as three spaces *inside* the black run, so a black rectangle appeared beside "Claude Code" — intermittently, because it only needs a re-sync to happen while the banner is on screen. Ghostty and VS Code never show it: they own the PTY and never re-serialise a screen.

## Fixed in the engine, not here

`termio-sh/libghostty-rs#3` carries the fix as a patch series against its pinned ghostty. ghostty's blank-*row* path already reset the style for exactly this reason; blank *cells* did not. Doing it there covers the viewport, the scrollback and a pending-wrap cursor at the source, at no runtime cost.

A repair in `termiod-vt` was written first and abandoned. Reading the grid back and erasing the bleed needs a `CUP` that **origin mode redirects onto live text** (turning `ABCDEFGHIJ` into `A___EFGHIJ`), and a `DECSC` slot an **already-attached client still holds across a resync** (a later `DECRC` landed at row 1 col 6 instead of row 3 col 7). Both were reproduced before the approach was dropped; it also missed scrollback and pending wrap entirely, and cost 0.37–1.64 ms per snapshot.

## Verification

- ghostty's own suite on the patched tree: **3411 pass, 26 skip, 0 fail**
- `cargo test -p termiod-vt` against the bumped pin: **34 pass, 1 ignored** (pre-existing reflow defect)
- `cargo build` clean

The five new tests replay bytes captured off a real Claude Code v2.1.246 PTY, and cover the reported gap, the alternate screen, a pending-wrap row and a row that has scrolled into history. They assert on background rather than whole cells: blank cells have always replayed as spaces rather than unwritten cells, which is invisible and predates this.

## Merge order

Depends on `termio-sh/libghostty-rs#3`. The pin here is the branch commit `05d888e`; re-point it at `master` once that merges.

Release Notes:

- Fixed a black rectangle that could appear next to an agent's banner after reattaching or resizing a session.


jiweiyuan:
Pin moved to `62e5f31` — termio-sh/libghostty-rs#3 needed two CI fixes after the patch series landed: Windows could not open a `canonicalize`d extended-length path through MSYS git, and the bindings-drift job reads `.ghostty-commit` to find the pin's build output, so the series digest moved to `.ghostty-patches` beside it.

jiweiyuan:
termio-sh/libghostty-rs#3 is merged; the pin now points at `0ef253c` on `master` rather than a branch commit.

PR #690, opened 2026-09-21
Judge a failed agent launch by the process's age

jiweiyuan:
Follow-up to #688. That PR added a floor so an agent exiting 0 the instant it launches parks on its output instead of silently handing the pane back to a shell. Reviewing the constant turned up two defects in it — one of them a false positive I introduced.

## The metric was wrong

The floor read `runtimeMilliseconds`, which `TermiodSessionLink` computes as `Date().timeIntervalSince(startedAt)` — and `startedAt` is set when the **link is constructed**, i.e. at attach, not at spawn.

Sessions outliving the app is the whole point of the daemon backend. So an agent running for hours that quits just after the app reattaches reported an age of milliseconds, and #688 would park it — on a session the user was entitled to get a shell back from.

`SessionInfo.created_unix` is the daemon's own spawn stamp and already rides along on the exit's final row. The fix takes the larger of that and the attach age: an attach can never precede a spawn, so the client's measure is a floor under the true age, never an overestimate. That also means a remote box whose clock runs ahead of this Mac's degrades to the attach age instead of reporting a negative one.

## The constant was undersized

Ghostty answers the same question with `abnormal-command-exit-runtime` (default **250 ms**), and its macOS branch counts a zero exit inside that window as abnormal for exactly the reason this does — so the mechanism is the one the engine we embed already uses.

But the two are not timing the same thing. Ghostty spawns `/bin/sh -c`; termio spawns the user's *login, interactive* shell so agent CLIs are found on the real `PATH`. Ten runs of `zsh -ilc true` on a development Mac:

```
3400  1940  5180  1990  700  910  930  940  800  1020   ms
```

All of that is startup files, and all of it sits inside the process's lifetime before the agent is reached. At 1 s the floor was inside that noise band — a genuinely failed launch frequently measured 0.7–2 s and sailed over it, so the net often did not fire at all. 10 s clears the worst measured startup with roughly 2× margin and stays far below any session someone actually worked in.

## What it still does not do

It stays a one-sided net, and the comment now says so rather than implying otherwise. A machine whose rc files take longer than the floor can still hide a failed launch, and no constant fixes that — only a signal separating "the shell was starting up" from "the agent was running" would, which today's exit event does not carry. The asymmetry is deliberate: a false park costs one keypress, a false revert costs the user an evening.

## Verification

- `swift build` clean; `swift test --filter TermiodStatusTests` 40 passed.
- Three new tests: the spawn stamp outranks the attach age, a missing/skewed stamp falls back, and a failure after a slow login shell is still caught.
- The false-positive test was verified to fail on the old metric — stubbing `processAgeMilliseconds` back to `sinceAttach` gives `("40") is less than ("3600000")` and `("park") is not equal to ("revertToShell")`.

Release Notes:

- Quitting an agent shortly after reopening Termio now hands the pane back to a shell, as quitting one always has, instead of leaving it on an exit prompt.


jiweiyuan:
Closing: the review found the premise wrong. `startedAt` is set at link construction, before `Transport.open` — so the attach age includes time *before* the spawn and is not a lower bound on the process's age, which is what the `max()` here rested on. Clock skew is also only guarded in one direction, and a park costs the pane rather than a keypress (`onClose` calls `closeSession`). Replaced by a PR that removes the floor and fixes the two daemon dependencies the review turned up instead.

